PR #2, opened 2015-01-31
Ned/add testing

nedbat:
Testing is cool.  @cpennington I'm sure you agree.


Makefile
```diff
@@ -0,0 +1,14 @@
...
+default: test
+
+test:
+	python test/test_func.py
+
+clean:
+	-rm -rf *.egg-info
+	-rm -f *.pyc */*.pyc */*/*.pyc */*/*/*.pyc */*/*/*/*.pyc */*/*/*/*/*.pyc
+	-rm -rf __pycache__ */__pycache__ */*/__pycache__ */*/*/__pycache__ */*/*/*/__pycache__ */*/*/*/*/__pycache__
+	-rm -f MANIFEST
```
cpennington: What does the `-` prefix mean?

cpennington:
I'm a bit baffled by how/why this works, but you appear to be invoking magic from the pylint test suite, so I'm happy to assume it's ok. We should hook up to travis, though.

Makefile
nedbat: Ignore errors: https://www.gnu.org/software/make/manual/html_node/Errors.html#Errors

nedbat:
TBH, this is mostly copy-paste from other test suites.  I looked a bit at the Pylint test runner.  The name of the .py file determines a number of things, including whether to expect pylint violations, and what versions of Python it applies to (though none of my tests use versions).  If the file is called "func_noerror_*.py", then we expect no violations.  Otherwise, the file is expected to generate the violations found in the same-named file in the messages folder, replacing ".py" with ".txt"

cpennington:
:

cpennington:
:+1:

PR #3, opened 2015-02-01
Ned/i18n checker

nedbat:
A checker for i18n translation functions without string literals.


nedbat:
@cpennington this checker has already found problems in edx-platform.  We should merge it :)

cpennington:
:+1:

PR #5, opened 2015-04-13
Sarina/ignore parseoption docstring

sarina:
@cpennington or @nedbat or @feanil

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,2 +1,3 @@
 Calen Pennington <calen.pennington@gmail.com>
 Ned Batchelder <ned@nedbatchelder.com>
+Sarina Canelake <sarina@edx.org>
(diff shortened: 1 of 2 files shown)


feanil:
:+1:  but how can I be sure without a travis build?

nedbat:
I don't know what `parse_options` is, why doesn't it deserve a docstring?  It's defined once in edx-platform, in capa/inputtypes.py:OptionInput.

sarina:
Its used in management commands to parse options given. It seems to have a
straightforward enough name ro not require a docstring. But if not I'll
just disable the warning explicitly or write stupid docstrings

nedbat:
Hmm, why do none of the management commands in edx-platform have this method? What drew your attention to it?

sarina:
Fixing up Carlos's code. If it's not a generic pattern, close this PR and
I'll write dumb docstrings.

nedbat:
My concern about this is that "parse_options" is kind of a generic name that is not necessarily a management command method.  The only current use of the name in edx-platform is not in a management command, for example.  I'd rather be conservative and require the docstring.

PR #11, opened 2015-09-11
Make it work on Python 3

nedbat:


clintonb:
:+1:

Confirmed changes work with Python 3.4.
Remember to update the version setup.py.

cpennington:
:+1:, and I'm going to push the big green button.

PR #12, opened 2015-09-18
Prepare package for distribution on PyPI

rlucioni:
I'd like to upload this package to the edX PyPI account. To prepare for that, I've made a few minor tweaks to the README and setup.py.

@nedbat @clintonb @jzoldak


clintonb:
:+1:

PR #13, opened 2015-09-29
New Checker for Renegade assertTrue/False

Sveder:
This is another Pylint checker that checks that comparisons aren't done inside of assertTrue/False, for example:
```
assertTrue(x == y) 
```
should be:
```
assertEqual(x, y)
```

I've started it here:
https://github.com/Sveder/custom-lints
and after email discussion with Ned Batchelder we agreed that it is a right fit to the edx-lint repo.


nedbat:
Thanks, I'm trying this out now.

Sveder:
@nedbat updated the PR - addressed your comments.

nedbat:
Great thanks!

PR #15, opened 2015-10-16
Allow "class Meta" with no docstring.

nedbat:
Also, pin the ends of the regex to be stricter.


nedbat:
@mattdrayer No need to use docstrings on "class Meta(object):", right?

nedbat:
@singingwolfboy ??

singingwolfboy:
seems reasonable to me. :+1:

PR #17, opened 2015-11-03
Use pylint-django

nedbat:
This reduces a lot of false negatives, especially for Django 1.8

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup
 
 setup(
     name='edx-lint',
-    version='0.3.1',
+    version='0.3.2',
     description='edX-authored pylint checkers',
     url='https://github.com/edx/edx-lint',
     author='edX',
@@ -30,6 +30,7 @@ setup(
     },
 
     install_requires=[
-        'pylint>=1.4.0',
+        'pylint==1.4.4',
+        'pylint-django==0.6.1',
     ],
 )
--- a/README.rst
+++ b/README.rst
@@ -36,8 +36,7 @@ Developing edx_lint
 
 To run the tests::
 
-    $ pip install -r requirements.txt
-    $ pip install -e .
+    $ pip install -r dev-requirements.txt
     $ make
 
 
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,0 +1,3 @@
+coverage
+six
+-e .
--- a/edx_lint/files/pylintrc
+++ b/edx_lint/files/pylintrc
@@ -19,7 +19,7 @@ persistent=yes
 
 # List of plugins (as comma separated values of python modules names) to load,
 # usually to register additional checkers.
-load-plugins=edx_lint.pylint
+load-plugins=edx_lint.pylint,pylint_django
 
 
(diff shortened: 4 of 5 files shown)


nedbat:
@singingwolfboy @jcdyer have a look-see.

setup.py
```diff
@@ -30,6 +30,7 @@
     },
 
     install_requires=[
-        'pylint>=1.4.0',
+        'pylint==1.4.4',
+        'pylint-django==0.6.1',
```
singingwolfboy: Shouldn't these changes be reflected in `requirements.txt`? Frankly, I'm not sure why this project has both a `requirements.txt` file *and* an `install_requires` section in the setup.py (that doesn't read the `requirements.txt` file).

setup.py
```diff
@@ -4,7 +4,7 @@
 
 setup(
     name='edx-lint',
-    version='0.3.0',
+    version='0.3.2',
```
singingwolfboy: Why the bump from `0.3.0` to `0.3.2`, rather than `0.3.1`? Not a problem, just curious for the rationale.

setup.py
jcdyer: Agreed that requirements and install_requires should be kept in sync.

nedbat: I've changed requirements.txt to not mention pylint, instead to install setup.py.

setup.py
nedbat: That was an indication that I needed to rebase: master was already at 0.3.1

setup.py
singingwolfboy: In that case, what's the purpose of the `requirements.txt` file? Can that file just be deleted?

nedbat: requirements.txt is for developers working on edx-lint.  Setup.py is for people using edx-lint.  I thought this was a well-established pattern?  I guess not.

nedbat: Maybe if I rename it to dev-requirements.txt, we'll be OK?

singingwolfboy: The pattern that I thought was the most well-established was:
* `requirements.txt` lists dependencies to use the software
* `setup.py` reads `requirements.txt` and passes the contents to the `install_requires` block
* If the software has additional dependencies for development, such as for running tests or building documentation, those dependencies are either listed in the `tests_require` section of `setup.py`, or listed in a secondary requirements file (such as `dev-requirements.txt`) which is similarly read into `setup.py`

Is there some documentation that we can refer to, to determine the correct pattern? I'm willing to be proven wrong.

jcdyer:
:+1:

setup.py
nedbat: There is no correct pattern, there are evolving best practices.  I don't see the advantage of setup.py reading a text file.  I'll rename the file to dev-requirements.txt.  The README will direct them to install from that if they are developing in the repo.  The setup.py will get the correct requirements for installed use.  OK?

singingwolfboy: That sounds good to me.

nedbat: 🎼 It's done, gimme some thumb 🎶

singingwolfboy:
:+1:

PR #19, opened 2015-11-06
Make edx-lint commands work with both python2 and python3

cpennington:
@nedbat, @clintonb: I had to make these changes to make it so that the IDA cookiecutter could run in both python2 and python3 virtualenvs.

### DIFF
--- a/edx_lint/tamper_evident.py
+++ b/edx_lint/tamper_evident.py
@@ -15,7 +15,7 @@ class TamperEvidentFile(object):
     def __init__(self, filename):
         self.filename = filename
 
-    def write(self, text, hashline=b"# {}"):
+    def write(self, text, hashline=u"# {}"):
         """
         Write `text` to the file.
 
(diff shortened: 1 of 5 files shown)


clintonb:
:+1:

edx_lint/tamper_evident.py
```diff
@@ -26,21 +26,21 @@ def write(self, text, hashline=b"# {}"):
         be changed to accommodate different file syntaxes.
 
         Arguments:
-            text (byte string): the contents of the file to write.
+            text (text string): the contents of the file to write.
```
nedbat: Let's get off on the right Python 3 foot, and say "unicode string", since that's what it is.

nedbat:
:+1: once the type names are changed in the docstring.

edx_lint/tamper_evident.py
cpennington: Ok. I tend to refer to the distinction as text vs bytes, and have seen others do the same, but I'm fine w/ unicode as well.

PR #20, opened 2015-11-23
Also check delattr for literal-used-as-attribute

nedbat:


nedbat:
@jcdyer @cpennington Turns out we had some of these in edx-platform.

cpennington:
:+1:

jcdyer:
:+1:

Issue #22, opened 2015-11-25
Setup Travis


clintonb:
- Automate tests
- Automate pushes to PyPI

nedbat:
I would like to divide this repo into the three separate repos it should be, rather than push this one to PyPI.

clintonb:
@nedbat do you have an ETA for that work?

nedbat:
I don't, it's all spare-time work.  I guess it isn't a disaster to push the three-headed beast to PyPI, we can push the proper repos later when we have them.

clintonb:
@nedbat I'm fine holding off on PyPI. I at least want to run tests. I prefer "build passed on Travis" to "It works on my machine".

PR #23, opened 2015-11-30
Upgraded to Pylint 1.5.x

clintonb:
- Updated Pylint requirement
- Removed obsolete configuration options

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import setup
 
 setup(
     name='edx-lint',
-    version='0.3.2',
+    version='0.4.0',
     description='edX-authored pylint checkers',
     url='https://github.com/edx/edx-lint',
     author='edX',
@@ -30,7 +30,7 @@ setup(
     },
 
     install_requires=[
-        'pylint==1.4.4',
-        'pylint-django==0.6.1',
+        'pylint>=1.5.1,<1.6.0',
+        'pylint-django>=0.7.1,<1.0.0',
     ],
 )
--- a/edx_lint/files/pylintrc
+++ b/edx_lint/files/pylintrc
@@ -7,9 +7,6 @@
 # pygtk.require().
 #init-hook=
 
-# Profiled execution.
-profile=no
-
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
 ignore=
@@ -84,10 +81,6 @@ reports=no
 # (RP0004).
 evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / statement) * 10)
 
-# Add a comment according to your evaluation note. This is used by the global
-# evaluation report (RP0004).
-comment=no
-
 
 # Template used to display messages. This is a python new-style format string
 # used to format the message information. See doc for all details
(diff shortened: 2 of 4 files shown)


clintonb:
Ideally, I'd like to resolve #22 before merging this.

setup.py
```diff
@@ -4,7 +4,7 @@
 
 setup(
     name='edx-lint',
-    version='0.3.2',
+    version='0.4.2',
```
nedbat: This should be 0.4.0, right?

setup.py
```diff
@@ -30,7 +30,7 @@
     },
 
     install_requires=[
-        'pylint==1.4.4',
+        'pylint>=1.5.0,<1.6.0',
```
nedbat: Currently, edx-platform only gets pylint installed because of this dependency.  Developers have complained in the past of the unpredictability of the differences between running pylint locally and on CI machines.  I think it would be best to keep this precisely pinned to reduce variation.

setup.py
clintonb: Fixed.

setup.py
clintonb: Pinned to 1.5.0.

clintonb: @nedbat this just caused problems for me (#24). If specific projects have issues with versions of pylint, those projects should pin the version. Tying this library to a specific version, however, is problematic for those who want to use other versions.

nedbat:
:+1:

clintonb:
I am going to merge this and create a separate PR to resolve #22. This is due to the fact that we cannot test Python 3.5 without Pylint>=1.5.1. Also, the Travis changes may require a deeper review.

dev-requirements.txt
```diff
@@ -1,3 +1,3 @@
 coverage
-six
+logilab-common
```
nedbat: Why is this here?

dev-requirements.txt
```diff
@@ -1,3 +1,3 @@
 coverage
-six
```
nedbat: Don't we still use six?

edx_lint/tamper_evident.py
```diff
@@ -53,18 +53,16 @@ def validate(self):
...
-        start_last_line = text.rfind(u"\n", 0, -1)
+        start_last_line = text.rfind("\n", 0, -1)
         if start_last_line == -1:
             return False
 
         original_text = text[:start_last_line+1]
         last_line = text[start_last_line+1:]
 
-        expected_hash = hashlib.sha1(original_text).hexdigest().encode("ascii")
+        expected_hash = hashlib.sha1(original_text.encode("utf-8")).hexdigest()
```
nedbat: This doesn't look right to me: on Py 2, original_text will be bytes already, so using .encode will raise an exception if there are non-ASCII characters.  My bad: there are no tests of non-ascii characters.

dev-requirements.txt
nedbat: Oops, six is in setup.py now.  Sorry.

edx_lint/tamper_evident.py
nedbat: Wait, the code explicitly restrict the contents to ascii? weird.

clintonb:
@nedbat Please comment on #26 regarding Travis, six, etc.

PR #26, opened 2015-12-02
Added Travis CI configuration

clintonb:


dev-requirements.txt
```diff
@@ -1,3 +1,3 @@
 coverage
-six
```
clintonb: `six` is actually needed by the package, not just for testing. Moved to setup.py.

dev-requirements.txt
```diff
@@ -1,3 +1,3 @@
...
+logilab-common
```
clintonb: I am guessing pylint 1.4.4 included this. pylint 1.5.1 does not.

clintonb:
@nedbat @cpennington please review. The changes were made to get tests passing for Python 3.5. I am not too familiar with bytes vs. str vs. unicode, so let me know if anything seems wonky.

Makefile
```diff
@@ -19,3 +19,6 @@ clean:
 	-rm -f *.pyc */*.pyc */*/*.pyc */*/*/*.pyc */*/*/*/*.pyc */*/*/*/*/*.pyc
 	-rm -rf __pycache__ */__pycache__ */*/__pycache__ */*/*/__pycache__ */*/*/*/__pycache__ */*/*/*/*/__pycache__
 	-rm -f MANIFEST
```
cpennington: How's about just `find . -name __pycache__ -delete'

test/test_tamper_evident.py
```diff
@@ -73,7 +73,7 @@ def test_editing_is_detected(self):
         self.assertFalse(tef.validate())
 
     def test_oneline_file_is_detected(self):
-        filename = self.write_tamper_evident(b"Am I OK?")
+        filename = self.write_tamper_evident("Am I OK?")
```
cpennington: All of the changes in this file are basically saying to use unicode strings in python 3, and bytestrings in python 2. That seems questionable to me. `write_tamper_evident` should expect one or the other.

Notably, the `TamperEvidentFile.write` method says that the string should be unicode, which it wasn't before (it was bytes before). That seems like it was a bug. I think it would be good to use `u` prefixed strings explicitly here, because that will force unicode strings in both 2 and 3.

nedbat: @cpennington: you change `TamperEvidentFile.write` from (implicitly) bytes to explicit Unicode strings when you made the code run on both 2 and 3.  That now seems like a bad choice of signature: file writing seems like a bytes-oriented operation to me.  But in any case, it should definitely be one or the other.

nedbat: Actually, wtf? I made changes in 6c728fb ("Make it work on Python 3") to change everything to explicit bytes.  The tests passed.  Then commit 0c34437 ("Make edx-lint commands work with both python2 and python3") changed it to explicit Unicode, and the tests fail.

@cpennington Why didn't we stick with explicit bytes?

cpennington: Gah... I'm not sure why I changed it to explicit unicode. Let me try and jog my memory, and I'll get back to you.

cpennington: So, looking at it, I think the reason I changed it to unicode was that it made more sense to me that way. In particular, it's talking about writing "text", and the explicit-bytes version made assumptions about the encoding of those bytes (although I suppose only in the hashline).

I know I also had difficulties with the interfaces for `ConfigParser`, which changed between 2 and 3 in how it handled unicode and strings. Not sure if that influenced my decision or not.

nedbat: I can fix it all up once we understand the pain points.

clintonb: Have you two identified the pain points? Are there still issues with bytes vs. unicode, or is this ready for final review?

nedbat:
The code as-is passes the tests on both 2 and 3, so I guess we can check it in.  I can take a pass over it later to get it onto a strong theoretical foundation.
:+1:

PR #28, opened 2015-12-16
Allow naming conventions up to 40 characters long

stvstnfrd:
While I've run into this wall previously, I was again prompted to take
action here by a recent PR [1] where a well-intentioned variable name
needed to be truncated just to pass quality checks.

This is one of those instances where the letter of the law (no long
names) conflicts with the spirt of the law (chose names that effectively
communicate purpose). As such, we should adapt.

> If we find the current quality configuration to be burdensome to the
> point of curtailing free expression, we should just update the config
> :)

[1] https://github.com/edx/edx-platform/pull/10756


nedbat:
@stvstnfrd I've created #29 to improve the docs.

For this pull request, you've edited both files, which isn't quite right, you need to generate the pylintrc from edx-lint/files/pylintrc.  There's a hash at the end of the file that won't be right unless you do.

nedbat:
This was handled in #30.

PR #30, opened 2015-12-18
Collected improvements requested by others.

nedbat:


nedbat:
@sarina @robrap @stvstnfrd @clintonb if you would....

clintonb:
:+1: (once tests pass)

nedbat:
@clintonb the tests aren't going to pass: we've enabled travis integration, but need your PR merged before we have a correct .travis.yml

pylintrc
```diff
@@ -179,4 +180,4 @@ int-import-graph =
 [EXCEPTIONS]
 overgeneral-exceptions = Exception
 
-# 3dd01da1adbb2cac2fbed3cbc578b3f647b792c7
```
sarina: what is this?

nedbat: Maybe I should add a comment here: "edx-lint write" includes a hash at the end of the file so that edits to the generated file will be detected in the future to prevent losing edits.  The hash will change any time the file contents change.

sarina: Cool. :+1:

clintonb:
@nedbat Oh. #26 merged.

robrap:
:+1:

nedbat:
@clintonb blah: pylint 1.5.1 still crashes edx-platform, so I had to go back to 1.4.4, which doesn't support Python 3.5 yet.

pylintrc
stvstnfrd: Sorry if I'm being dense, but I'm not sure I understand this:

>  "edx-lint write" includes a hash at the end of the file so that edits to the generated file will be detected in the future to prevent losing edits

Aren't `pylintrc` files already under version control? What's the use-case that would cause a loss of changes?

stvstnfrd:
Just a tangential question, but still +1

pylintrc
nedbat: With this tool in place, a pylintrc file is not a source file, it's generated from a central pylintrc, and a local pylintrc_tweaks.  If you edit the pylintrc file directly, and then later someone overwrites it with "edx_lint write pylintrc", then your edited changes will be lost.  The hash allows edx_lint to detect that the file had been edited since generation, and keep the edited file for you to resolve by hand.

PR #31, opened 2015-12-20
Clean up, especially for Python 3

nedbat:
This changes the file operations back to pure byte strings.  Also, uses tox for testing, so we can easily test Python 2 and 3 locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ test:
 	tox -e py27,py35
 
 pylint:
-	pylint edx_lint test setup.py
+	tox -e pylint
 
 clean:
 	-rm -rf *.egg-info
@@ -16,6 +16,7 @@ clean:
 	-find . -name '*.pyc' -delete
 	-rm -f MANIFEST
 	-rm -rf .coverage .coverage.* htmlcov
+	-rm -rf .tox
 
 requirements:
 	pip install -r dev-requirements.txt
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,10 @@ pylint:
 
 clean:
 	-rm -rf *.egg-info
-	-rm -f *.pyc */*.pyc */*/*.pyc */*/*/*.pyc */*/*/*/*.pyc */*/*/*/*/*.pyc
-	-rm -rf __pycache__ */__pycache__ */*/__pycache__ */*/*/__pycache__ */*/*/*/__pycache__ */*/*/*/*/__pycache__
+	-find . -name '__pycache__' -prune -exec rm -rf "{}" \;
+	-find . -name '*.pyc' -delete
 	-rm -f MANIFEST
+	-rm -rf .coverage .coverage.* htmlcov
 
 requirements:
 	pip install -r dev-requirements.txt
--- a/Makefile
+++ b/Makefile
@@ -5,11 +5,7 @@
 default: test
 
 test:
-	python -m unittest discover
-
-coverage:
-	coverage run -m unittest discover
-	coverage report -m
+	tox -e py27,py35
 
 pylint:
 	pylint edx_lint test setup.py
--- a/edx_lint/tamper_evident.py
+++ b/edx_lint/tamper_evident.py
@@ -15,7 +15,7 @@ class TamperEvidentFile(object):
     def __init__(self, filename):
         self.filename = filename
 
-    def write(self, text, hashline=u"# {}"):
+    def write(self, text, hashline=b"# {}"):
         """
         Write `text` to the file.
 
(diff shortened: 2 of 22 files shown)


nedbat:
@cpennington @clintonb This is how I think we should support both 2 and 3.  Let me know if something goes wrong.  Also, I put tox in here so we can test both Python versions locally without multiple virtualenvs.

Makefile
```diff
@@ -5,20 +5,17 @@
 default: test
 
 test:
-	python -m unittest discover
-
-coverage:
-	coverage run -m unittest discover
-	coverage report -m
+	tox -e py27,py34
```
clintonb: `py35`? We are using Python 3.5 for https://github.com/edx/course-discovery.

Makefile
```diff
@@ -5,20 +5,17 @@
...
 
 pylint:
 	pylint edx_lint test setup.py
 
 clean:
 	-rm -rf *.egg-info
-	-rm -f *.pyc */*.pyc */*/*.pyc */*/*/*.pyc */*/*/*/*.pyc */*/*/*/*/*.pyc
-	-rm -rf __pycache__ */__pycache__ */*/__pycache__ */*/*/__pycache__ */*/*/*/__pycache__ */*/*/*/*/__pycache__
+	-find . -name '*.pyc' -delete
+	-find . -name '__pycache__' -delete
```
clintonb: I was looking into this Friday morning. This will fail if the directory is not empty.

clintonb:
:+1: pending ...

- [ ] Python 3.5 support
- [ ] Updated clean target

Makefile
nedbat: Pylint 1.4.4 doesn't support Python 3.5, and Pylint 1.5 crashes on edx-platform.

Makefile
nedbat: Really? It seems to work for me on Mac.

nedbat: Oh, I know why: because if the only files in there are .pyc files, then they were all deleted in the previous step :)
I'll make a better command.

nedbat:
Now updated to the newly released Pylint 1.5.2, and Python 3.5.  Doesn't crash edx-platform, but increases the violations by about 4000!

Makefile
cpennington: But `.travis` is using `3.5`.

edx_lint/tamper_evident.py
```diff
@@ -26,21 +26,21 @@ def write(self, text, hashline=u"# {}"):
         be changed to accommodate different file syntaxes.
 
         Arguments:
-            text (unicode string): the contents of the file to write.
+            text (byte string): the contents of the file to write.
 
-            hashline (unicode string): the format of the last line to append to
+            hashline (byte string): the format of the last line to append to
                 the file, with "{}" replaced with the hash.
```
cpennington: Please add that this must be `utf-8` encoded.

cpennington: In fact, both arguments must be, otherwise the file contents will be encoded with different encodings in different places, which would be unfortunate.

cpennington: (I think that's why I made this take unicode when I changed it)...

edx_lint/tamper_evident.py
```diff
@@ -50,19 +50,19 @@ def validate(self):
...
+        start_last_line = text.rfind(b"\n", 0, -1)
         if start_last_line == -1:
             return False
 
         original_text = text[:start_last_line+1]
         last_line = text[start_last_line+1:]
 
-        expected_hash = hashlib.sha1(original_text.encode("utf-8")).hexdigest()
-        match = re.search(r"[0-9a-f]{40}", last_line)
+        expected_hash = hashlib.sha1(original_text).hexdigest().encode('ascii')
```
cpennington: This encoding differs than the `actual_hash` encoding used above (not materially, I suppose, since the hexdigest is all ascii... but still, consistency...)

test/test_pylint_plugins.py
```diff
@@ -13,7 +11,6 @@ def load_tests(loader, tests, pattern):
 
     # Load our plugin, and disable messages we don't want noising up our tests.
```
cpennington: Fix this comment, too?

Makefile
cpennington: And `.tox` only has a `py35` environment.

nedbat: Fixed.

edx_lint/tamper_evident.py
nedbat: It's really stupid that the only reason we care about which encoding is because bytestrings don't have a .format() method... :(   In any case, requiring UTF8 is fine with me.

nedbat:
@cpennington fixed everything.

PR #33, opened 2015-12-21
Improved assertion checker

nedbat:


nedbat:
@jcdyer @cpennington Better messages about assertions that can be improved.

edx_lint/pylint/right_assert_check.py
```diff
@@ -28,6 +28,40 @@ class AssertChecker(BaseChecker):
...
+        ("assertLess", "assertGreaterEqual"),
+        ("assertGreater", "assertLessEqual"),
+        ("assertIs", "assertIsNot"),
+        ("assertIsNone", "assertIsNotNone"),
+    ]
+
+    INVERTED = {}
+    for yup, nope in INVERTED_PAIRS:
+        INVERTED[yup] = nope
+        INVERTED[nope] = yup
```
cpennington: :+1: just for the variable names. =)

edx_lint/pylint/right_assert_check.py
```diff
@@ -51,39 +85,21 @@ def visit_callfunc(self, node):
...
-                node=node,
-            )
-
-        elif "<" in first_arg.ops[0][0] or ">" in first_arg.ops[0][0]:
-            # An assertTrue/False with a comparison should be assertGreater or assertLess:
+        existing_code = "%s(%s)" % (node.func.attrname, first_arg.as_string())
+
+        if isinstance(first_arg, astroid.Compare):
+            compare = first_arg.ops[0][0]
+            right = first_arg.ops[0][1]
```
cpennington: Is it worth checking both `right` and `left` (I'm assuming those are the arguments to the comparison, and so `None` could be either one?)

nedbat: I wondered about that.  In fact, because of chained comparisons, the structure of a Compare node is that `left` is an expression, and then `ops` is a list of pairs, (compare_op, expression).  So `a == b == c <= d` is expressed as `(left=a, ops=[("==", b), ("==", c), ("<=", d)])`.  We don't check here if there is more than one comparison either.

cpennington:
:+1:

PR #37, opened 2016-03-09
Additional clarity for pylintrc_tweaks how-to.

benpatterson:
@nedbat @clintonb @cpennington

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,13 +24,22 @@ edx_lint master file::
 
     $ edx_lint write pylintrc
 
-You can customize the resulting file by creating a pylintrc_tweaks file in the
-current directory.  It should contain only the settings you want to override.
-
 The file written contains a hash of its contents, to detect subsequent editing.
 ``edx_lint`` will detect this when it next tries to write the file.  If editing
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -24,14 +24,23 @@ edx_lint master file::
...
 is detected, the edited file will be moved aside so it can be compared to the
 newly written file.
 
+Customizing edx_lint
+--------------------
+You can customize the resulting pylintrc file by creating a pylintrc_tweaks file in the
+current directory before running the ``write`` sub-command.  It should contain only the
+settings you want to override.
+
+**Note:** If your project is not a django project, you'll want to disable the django plugins with
```
clintonb: Capitalize Django.

clintonb:
I'm still not sure it makes sense to include the Django plugin(s) by default, since this isn't only intended for use with Django. However, we can hash that out later.

:+1:

README.rst
benpatterson: Done...amended commit & re-pushed.

nedbat:
I agree that this is an unfortunate way to solve the problem.  I'll figure out a better way, but in the meantime, :+1:

PR #47, opened 2016-09-23
fix typo in warning comment

pwnage101:
the generated pylintrc for this repo checks out:
```
% python -m edx_lint check pylintrc
Checking existing copy of pylintrc
Your copy of pylintrc is good
```


nedbat:
:+1: Merge it!

pwnage101:
@nedbat I just changed this from an FYI PR to an actual PR.  Just learned about the hash and decided to regenerate this repo's pylintrc.

nedbat:
What is an FYI pr?

pwnage101:
Ah, I was just going to push the PR without seeking review because it seemed to straightforward (known as an FYI PR (as far as I know).  Now it isn't seeming so straightforward to me because of things that get hashed.

nedbat:
The hash is fine, and the fix is correct, I'm merging it :)

PR #49, opened 2016-10-05
Polish up layered_test_check

nedbat:
Move method out of the class that could be useful for future thing.  Booleanify the return value just in case. Add test cases to get coverage to 100%

### DIFF
--- a/edx_lint/pylint/layered_test_check.py
+++ b/edx_lint/pylint/layered_test_check.py
@@ -14,6 +14,26 @@ def register_checkers(linter):
     linter.register_checker(LayeredTestClassChecker(linter))
 
 
+def is_test_case_class(node):
+    """Is this node a test class?
+
+    To be a test class, it has to derive from unittest.TestCase, and not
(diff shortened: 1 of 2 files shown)


nedbat:
@cpennington

edx_lint/pylint/layered_test_check.py
```diff
@@ -32,32 +52,13 @@ class LayeredTestClassChecker(BaseChecker):
     }
 
     @utils.check_messages(MESSAGE_ID)
```
cpennington: Do we need to leave this decorator on the `is_test_case_class` method?

nedbat: No, in fact, that was one of the first things I changed.  That decorator is used on visit_ methods to limit the exploration. Not sure why we had it on a helper method originally.

nedbat: I should not assume that silence is assent, correct?

PR #50, opened 2016-11-02
Required base classes

nedbat:
This adds a new checker, for required base classes.  A new setting, `required-base-class`, is a list of colon-separated pairs:
```
required-base-class: unittest.case.TestCase:common.tests.helpers.TestMixin, ....
```
This says that if a class derives from TestCase, then it must also derive from TestMixin.  The names must be fully qualified class names.  This is unfortunate in some cases, since like with unittest.case.TestCase, the fully qualified name does not correspond to how it was imported.

### DIFF
--- a/edx_lint/pylint/required_base_class.py
+++ b/edx_lint/pylint/required_base_class.py
@@ -2,9 +2,6 @@
 
 import collections
 
-import astroid
-from astroid.scoped_nodes import get_locals     # not sure this is the right import
-
 from pylint.checkers import BaseChecker, utils
 from pylint.interfaces import IAstroidChecker
 
@@ -47,6 +44,7 @@ class RequiredBaseClassChecker(BaseChecker):
         self.class_map = collections.defaultdict(set)
 
     def open(self):
+        # pylint: disable=no-member
         if self.config.required_base_class:
             for pair in self.config.required_base_class:
                 child, parent = pair.split(":")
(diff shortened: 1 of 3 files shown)


nedbat:
@cpennington FYI

nedbat:
@cpennington This works now, and is ready for your review.  Ideas welcome about ways to improve the setting, since fully qualified names are a little awkward.

edx_lint/pylint/required_base_class.py
```diff
@@ -0,0 +1,74 @@
...
+                child, parent = pair.split(":")
+                self.class_map[child].add(parent)
+
+    @utils.check_messages(MESSAGE_ID)
+    def visit_class(self, node):
+        """Check each class."""
+        if not self.class_map:
+            return
+
+        all_bases = [usable_class_name(c) for c in node.mro()]
```
PCManticore: You should protect against NotImplementedError when calling .mro(), in case you encounter old style classes, for which the operation is undefined (we will going to raise a different exception in the future astroid 2.0 though).

test/test_required_base_class.py
```diff
@@ -0,0 +1,69 @@
...
+
+    CHECKER_CLASS = RequiredBaseClassChecker
+
+    def get_class_node(self, code):
+        """Parse `code`, and return the last class node.
+
+        The code should have at least one class definition.
+
+        """
+        node = astroid.parse(code)
```
PCManticore: Maybe you can use astroid.testutils.extract_node / astroid.extract_node for this? It looks something as in:

```
from astroid import extract_node
node = extract_node('''
class A:
    pass
class B: #@
   pass
class C:
   pass
''')
```

PR #53, opened 2017-03-07
Cale/lint amnesty

cpennington:


robrap:
@cpennington Reminder to add an amnesty note where it is disabling so we know which were intentional.  Thanks!

edx_lint/cmd/amnesty.py
```diff
@@ -0,0 +1,110 @@
...
+        return ""
+
+
+def fix_pylint(line, errors):
+    """
+    Yield any modified versions of ``line`` needed to address the errors in ``errors``.
+    """
+    if not errors:
+        yield line
+        return
```
cpennington: Actually, in this case, the return is so that the function ends early (otherwise, it would just continue after the first yield).

edx_lint/cmd/amnesty.py
```diff
@@ -0,0 +1,110 @@
...
+import click
+
+PYLINT_PARSEABLE_REGEX = re.compile(
+    r"""^(?P<filename>[^:]+):(?P<line>\d+): """
+    r"""\[(?P<error_number>[^(]+)\((?P<error_code>[^)]+)\), (?P<function>[^\]]*)\] """
+    r"""(?P<error_msg>.*)"""
+)
+PYLINT_EXCEPTION_REGEX = re.compile(r"""\s*#\s*pylint:\s+disable=(?P<disables>[^#]*)""")
+
+PylintError = namedtuple('PylintError', ['filename', 'line', 'error_number', 'error_code', 'function', 'error_msg'])
```
nedbat: Here "line" means line number, later in fix_pylint, it means the text of the line. "linenum" here would be good.

edx_lint/cmd/amnesty.py
```diff
@@ -0,0 +1,110 @@
...
+    Add ``# pylint: disable`` clauses to add exceptions to all existing pylint errors in a codebase.
+    """
+    errors = defaultdict(lambda: defaultdict(set))
+    for pylint_error in parse_pylint_output(pylint_output):
+        errors[pylint_error.filename][pylint_error.line].add(pylint_error)
+
+    for file_with_errors in sorted(errors):
+        with open(file_with_errors) as input_file:
+            output_lines = []
+            for line_num, line in enumerate(input_file):
```
nedbat: `enumerate(input_file, start=1)` was made for this.

edx_lint/cmd/amnesty.py
```diff
@@ -0,0 +1,110 @@
+"""
```
nedbat: No tests for this code?

edx_lint/cmd/amnesty.py
jzoldak: ah, thx for the explanation. learned something new!

edx_lint/cmd/amnesty.py
cpennington: Ohh, TIL.

edx_lint/files/.editorconfig
```diff
@@ -0,0 +1,25 @@
+# This is a file to standardize editor settings: http://EditorConfig.org
```
cpennington: I'd love to make it as authoritative and comprehensive as possible.

edx_lint/files/.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Matches multiple files with brace expansion notation
+# Set default charset
+[*.{js,py}]
+charset = utf-8
```
cpennington: Yes, I'd like to. Although, per your in-person suggestion, maybe I'll just pull the editorconfig out into a separate repo.

edx_lint/cmd/amnesty.py
cpennington: tests added.

cpennington:
There are a couple of `trailing_comment` tests that already exist.

cpennington:
@jzoldak: I've addressed your comments.

pylintrc
```diff
@@ -55,32 +55,293 @@
...
+	bad-continuation,
+	invalid-name,
+	misplaced-comparison-constant,
+	too-many-lines,
+	file-ignored,
+	too-many-nested-blocks,
+	too-many-statements,
+	too-many-boolean-expressions,
+	exec-used,
+	eval-used,
```
ormsbee: Could we please move `exec-used` and `eval-used` to the enabled list? I feel like those are rare and dangerous enough that it's worth automatically flagging.

cpennington: sure.

edx_lint/files/pylintrc
```diff
@@ -36,28 +249,93 @@ load-plugins=edx_lint.pylint,pylint_django,pylint_celery
...
 # --disable=W"
 disable=
+    # These should be left to the discretion of the reviewer
+    bad-continuation,
+    invalid-name,
+    misplaced-comparison-constant,
+    too-many-lines,
+    file-ignored,
+    too-many-nested-blocks,
+    too-many-statements,
```
benpatterson: I think this should default to enable; it calls out overly-complex code that can yield more bugs and maintainability problems.

pylintrc
```diff
@@ -55,32 +55,294 @@
...
+	bad-inline-option,
+	deprecated-pragma,
 disable = 
+	bad-continuation,
+	invalid-name,
+	misplaced-comparison-constant,
+	too-many-lines,
+	file-ignored,
+	too-many-nested-blocks,
+	too-many-statements,
```
benpatterson: Same here. This is a code complexity problem and I think it should be enabled by default.

cpennington: I feel pretty strongly that this is something we should be relying on our reviewers for. Anything that's long enough for pylint to complain about, I'd expect a reviewer to flag, unless there was a good reason.

ormsbee: Since the explicit goal of this exercise is to filter out noise, I agree with @cpennington that we should confine it to things that would warn of acute danger, and not general complexity issues.

benpatterson: Just the nature of a code review, with a window of a couple lines above and below, means that it will be easy for a method to add more and more over time, and reach a point in which it's become too big & complex, without anyone noticing.

That though, is my final volley. If you still think it should stay disabled, then carry on.

ormsbee: I agree that's a problem, but by that point it almost never makes sense for the implementer to do a major refactoring. Adding one more argument or line to the ugly-but-stable morass is usually far quicker and safer than trying to refactor it, since the latter is often an order of magnitude harder than the feature or bug they're working on in the first place. So the only thing people will do is slap on an ignore pragma and carry on.

The only time stuff like that gets truly fixed is when people have decided that this subsystem is painful enough to replace or revamp in its own story or epic.

benpatterson: @ormsbee thanks for the follow-up.

edx_lint/files/pylintrc
```diff
@@ -36,28 +249,93 @@ load-plugins=edx_lint.pylint,pylint_django,pylint_celery
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 disable=
+    # These should be left to the discretion of the reviewer
```
jmbowman: Do we have statistics on how often each of these checks is either violated or locally disabled?  And is there a significant increase in runtime from enabling them?

cpennington: This is less about increase/decrease in running speed, and more about checks that we frequently add exceptions to.

edx_lint/cmd/amnesty.py
```diff
@@ -0,0 +1,116 @@
...
+    if current:
+        original_errors = {disable.strip() for disable in current.group('disables').split(',')}
+    else:
+        original_errors = set()
+
+    disabled_errors = set(original_errors)
+
+    for error in errors:
+        if error.error_name == 'useless-suppression':
+            parsed = re.search("""Useless suppression of '(?P<error_name>[^']+)'""", error.error_msg)
```
cpennington: Mostly, I pre-compiled them to make it easier to use them more than once. The pre-compilation isn't actually required for any kind of performance impact.

test/test_amnesty.py
```diff
@@ -0,0 +1,297 @@
...
+            "some line# pylint: disable=some-exception#noqa",
+        )
+
+    def test_parse_pylint_output(self):
+        pylint_output = textwrap.dedent("""\
+            ************* Module cms.urls
+            cms/urls.py:1: [C0111(missing-docstring), ] Missing module docstring
+            ************* Module cms.envs.test
+            cms/envs/test.py:81: [C0301(line-too-long), ] Line too long (144/120)
+        """)
```
jzoldak: Please verify a pylint violation whose parseable output spans multiple lines. 
E.g. if I take out the spaces around the equals on line 74 above, I get the following output.
```
************* Module test.test_amnesty
test/test_amnesty.py:78: [C0326(bad-whitespace), ] Exactly one space required around assignment
        pylint_output=textwrap.dedent("""\
                     ^
```

Also we should verify parsing with multiple violations per file.

edx_lint/cmd/amnesty.py
```diff
@@ -0,0 +1,116 @@
...
+    errors = defaultdict(lambda: defaultdict(set))
+    for pylint_error in parse_pylint_output(pylint_output):
+        errors[pylint_error.filename][pylint_error.linenum].add(pylint_error)
+
+    for file_with_errors in sorted(errors):
+        with open(file_with_errors) as input_file:
+            output_lines = []
+            for line_num, line in enumerate(input_file, start=1):
+                output_lines.extend(
+                    fix_pylint(
```
jzoldak: Do we care if adding the pylint pragmas (plus whatever comment might already have been on that line) makes the line too long?

cpennington: No (although it does mean that you might need to run pylint again to squash the line-too-long errors).

test/test_amnesty.py
cpennington: Ooh, nice, I had forgotten about those errors.

test/test_amnesty.py
```diff
@@ -0,0 +1,297 @@
...
+        )
+        self.assert_pylint_exception_match(
+            "  # pylint: disable=some-exception",
+            "some line  # pylint: disable=some-exception  #noqa",
+        )
+        self.assert_pylint_exception_match(
+            "# pylint: disable=some-exception",
+            "some line# pylint: disable=some-exception#noqa",
+        )
+
```
cpennington: Will do.

edx_lint/cmd/amnesty.py
robrap: Seems like it should take too much code to determine if it would be newly needed.  Otherwise, instructions will need to include running it twice?

cpennington: I'll double-check and see if it complains if the line is too long due to pragmas.

robrap: Ok.  Remember to include your 'amnesty' comment in your test.  Thanks.

cpennington: mm... testing is good. Turns out, pylint doesn't like the trailing `# lint-amnesty` comment.

cpennington: OK, I verified that pylint doesn't count comments in line length (or at least, not the comments added by this module).

README.rst
```diff
@@ -29,6 +29,26 @@ The file written contains a hash of its contents, to detect subsequent editing.
 is detected, the edited file will be moved aside so it can be compared to the
 newly written file.
 
+Using lint-amnesty
+------------------
+
+The ``lint-amnesty`` command can be used to squash all existing pylint errors
+in a codebase, so that from then the repository can maintain pylint-cleanliness.
```
clintonb: ...from then on...

cahrens:
I think we should stop complaining about missing doc strings. Doc strings can be very helpful, but our code is also littered with doc strings that do nothing more than duplicate the self-documenting method name. I would like to leave it to reviewers to enforce documentation on methods that are complicated enough to warrant them.

cpennington:
@cahrens: I'm not sure if pylint can actually support this, but how would you feel about still requiring docstrings on functions and methods, but not on classes (and perhaps not on modules). I feel like classes and modules are the ones that are the most prone to ending up with content-free docstrings.

cahrens:
@cpennington that would be OK as long as methods in test classes don't require doc strings. That's the case I most often hit where the doc string is completely unnecessary because the test method name describes what the test is covering.

cpennington:
@cahrens: Yes, I agree that `test_*` methods should exempt (although custom assertion methods, or test setup helper methods probably shouldn't be)

robrap:
FYI: @nasthagiri @jcdyer: Since code comments came up as part of our Clean Code book review.

nedbat:
Test methods are already exempt from needing docstrings:
```
no-docstring-rgx = __.*__$|test_.+|setUp$|setUpClass$|tearDown$|tearDownClass$|Meta$
```
Seems like we could use better education about what we already do.

edx_lint/files/pylintrc
```diff
@@ -36,28 +249,88 @@ load-plugins=edx_lint.pylint,pylint_django,pylint_celery
...
+    bad-continuation,
+    invalid-name,
+    misplaced-comparison-constant,
+    file-ignored,
+    too-many-nested-blocks,
+    too-many-statements,
```
nasthagiri: `too-many-statements` seems like a good one to keep.  It violates our Clean Code initiative.  And if we ever run into it, it's a good time to refactor the function.  (If anything, the number of acceptable lines in a method could be reduced.)

cahrens:
Could we not require doc strings on inner methods? There's nothing more annoying than having to write a doc string on a 3-line helper method that you want to use throughout a method (and with our style convention of no single-line doc strings, you have now doubled the helper method length). It's not like anyone is ever going to rely on that doc string-- they are just going to read the implementation of the method.

I'm leaning towards not requiring doc strings ever. Documentation was useful in Java, where we often read doc separately from source code. But with Python (at least our Python), I am always going to look at the code to figure out what it is actually doing. Doc strings do hint at intent, but they are rarely (IMHO) useful. The times where doc strings are helpful are when they actually describe input arguments and what will be returned in various scenarios, but I seldom see doc with that level of detail.

edx_lint/files/pylintrc
```diff
@@ -36,28 +249,88 @@ load-plugins=edx_lint.pylint,pylint_django,pylint_celery
...
+    too-many-boolean-expressions,
+    protected-access,
```
nasthagiri: Can we disable `protected-access` for tests only?  In non-test code, this violation indicates that we're breaking an interface abstraction - or the interface needs to be updated.

edx_lint/files/pylintrc
```diff
@@ -36,28 +249,88 @@ load-plugins=edx_lint.pylint,pylint_django,pylint_celery
...
+    bad-indentation,
+    lowercase-l-suffix,
+    wildcard-import,
+    unused-wildcard-import,
+    global-statement
```
nasthagiri: Hmm.. `global-statement` seems like one that could prevent accidental programming errors, no?  Why disable this?

cpennington: No, this is specifically referring to the `global some_var` statement that introduces a global variable into a local scope.

edx_lint/files/pylintrc
cpennington: I'm not sure pylint will let us be that specific. And per my comment, this is something that I'd like for reviewers make the call on. That said, we could leave it on, but just give blanket permission to allow `_foo` access in test files (using a file-level `pylint: disable`).

edx_lint/files/pylintrc
cpennington: There was some discussion of this up above (@benpatterson had similar concerns, and @ormsbee and I responded)

cpennington:
@cahrens: Argument docs are exactly the reason that I want to require docstrings on functions and methods (and don't much care about classes and modules).

I suspect we can't exclude inner functions, sadly. I agree that the docstrings on those are usually not that useful.

pylintrc
nasthagiri: I agree with @benpatterson here.  This is a code complexity issue and should make the contributor stop-and-think before proceeding. 

Because of how git shows code differences, the entire function is not always visible to code reviewers.  I'd rather have the linter continue to flag this issue and then let the reviewer and contributor decide whether or not to slap on a pragma.

Also, I believe it is important to keep this listed as a linting issue that edX believes is important for code quality.  Having existing large functions in our code base is a matter that we should address and reduce, not begin to ignore due to concerns of productivity - especially since future productivity will continue to only decrease with larger and larger functions.

edx_lint/files/pylintrc
nasthagiri: sorry - missed it.  responded above.

edx_lint/files/pylintrc
nasthagiri: Meaning, put that top-level pragma on each test file in the amnesty?  And have developers do the same for new test files?

pylintrc
cpennington: You make a good point. One of the things that this whole lint-amnesty project is for is to make it so we don't have to atone for our past sins right away. So, we'll flag existing `too-many-statements` methods as being allow, but won't allow new ones.

ormsbee:
@cpennington: I don't know if it's easy to exclude, but another common place I find writing useless docstrings are dunder methods.

edx_lint/files/pylintrc
```diff
@@ -36,28 +249,88 @@ load-plugins=edx_lint.pylint,pylint_django,pylint_celery
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 disable=
+    # These should be left to the discretion of the reviewer
+    bad-continuation,
```
jcdyer: Is this one we regularly suppress?  It seems like it should be pretty straightforward to appease.

edx_lint/files/pylintrc
```diff
@@ -36,28 +249,88 @@ load-plugins=edx_lint.pylint,pylint_django,pylint_celery
...
+    invalid-name,
+    misplaced-comparison-constant,
+    file-ignored,
+    too-many-nested-blocks,
+    too-many-statements,
+    too-many-boolean-expressions,
+    protected-access,
+    bad-indentation,
+    lowercase-l-suffix,
+    wildcard-import,
```
jcdyer: Why do we want to start allowing starports?  This seems like a bad idea.

edx_lint/files/pylintrc
cpennington: It's fiddly, and not worth the hassle to appease.

cpennington:
@cahrens: I made it so that functions with < 5 lines won't require docstrings.
@ormsbee: Our pylintrc already excludes (or attempts to exclude) `__.*__` methods. Do you have any examples of being forced to document those (in a project that uses edx-lint)?

cpennington:
@nasthagiri, @jzoldak: Do you have any blockers?

ormsbee:
@cpennington: I thought I ran into it on edx-platform, but maybe I'm misremembering it. I'll open a PR here if I run into it again and it turns out to be a bug somehow.

nasthagiri:
No blockers from me either.  
Does the pylintrc file need to be regenerated?
Thanks for doing this!

nedbat:
This seems like a strange way to amend the coding standards. 1) People don't know these decisions are being made, and so can't participate. 2) We aren't updating the actual standards.

On the docstring issue: do we really feel like the current reviews are getting us good docstrings, and therefore will ensure we have docstrings on the functions that need them?

cpennington:
@nedbat: I attempted to inform the organization (via an all-hands). I admit, I didn't engage the larger OpenEdX community. Are there others that you feel should have weighed in?

Re: the actual standards: I assume you mean the OpenEdX docs?

Re: docstrings: as merged, we are still requiring docstrings on all things except functions <= 5 lines long.

cpennington:
@nedbat: After thinking a bit about this this morning, I think it makes sense to backstop these changes with an OEP. I'll create one soonish.

PR #56, opened 2017-04-19
Pin pylint and astroid versions

cpennington:


edx_lint/cmd/amnesty.py
```diff
@@ -58,8 +58,8 @@ def format_pylint_disables(error_names, tag=True):
             disabled=", ".join(sorted(error_names)),
             tag=tag_str,
         )
-    else:
-        return ""
+
+    return ""
```
cpennington: TBH, I don't really like it either. I'll disable the rule, and undo this change.

test/test_pylint_plugins.py
```diff
@@ -13,7 +13,49 @@ def load_tests(unused_loader, tests, unused_pattern):
     """
     # Have to import this in the function, because the module does
     # initialization on import! ugh.
-    from pylint.testutils import make_tests, LintTestUsingFile, cb_test_gen, linter
+    from pylint.testutils import LintTestUsingFile, linter, get_tests_info
+
+    # Inline functions that was deleted from pylint core
```
cpennington: Yeah, I didn't want to put in the work to try and figure out how their test suite works now.

PR #57, opened 2017-12-15
Stop pbr from making files, and ignore .eggs

nedbat:


nedbat:
@cpennington Help me hate pbr less :)

PR #58, opened 2018-01-10
Update .js files to use 2-space indents

cpennington:


nedbat:
@cpennington is this generally agreed-on?  I thought .js was a more complicated situation?

cpennington:
Our eslint settings seem to want 2-space indents. @arizzitano: is there more complexity than that?

PR #59, opened 2018-01-10
Do not require docstrings for private methods

jzoldak:


nedbat:
I don't understand why private methods don't need documentation?

jzoldak:
The intersection of requiring docstrings and writing small functions (e.g. as advocated in "Clean Code by Robert C. Martin", which has been making the rounds at edX. Here's a quote "The first rule of functions is that they should be small. The second rule of functions is that they should be smaller than that.") is leading to gratuitous docstrings that clutter the codebase without being helpful.

E.g.
def _do_foo(bar):
    """
    Do foo to bar.
    Return boof.
    """
    baz = _another_method_that_is_well_named_so_you_know_what_it_does(bar)
    boof = _another_other_well_named_method(baz)   
    return boof

nedbat:
What type is bar? What values are allowed? What type is boof? What can I expect from it?

I find that I often understand more about the code I've written when I have to explain it in English.  Taking the time to think about what the function *actually does* is very helpful.  People in the future will definitely have questions.  Answer them now.

jzoldak:
Our current pylint settings don't check anything about the content of the docstring such as information about types, values, and expectations. Call me jaded, but I don't believe that **requiring** a docstring increases the likelihood of developers writing **good** docstrings in the cases that they would be helpful. I think that devs are going to use whatever their judgement is at that time.

I'm not advocating dropping the requirement for public methods, and for those I personally agree with [our best practice guidelines when writing docstrings](http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#docstrings).

However, especially because we dictate the multi-line convention even if there’s only one line of content, I question the value of requiring docstrings for short private functions, and posit that this requirement is detrimental to code readability.

nedbat:
1. Your change is to remove a requirement for underscore-named ("private") functions, but your justification is all about short obvious functions.  I don't think private functions are necessarily short and obvious, and I don't think short obvious functions are necessarily private.

2. We already have a way to avoid writing docstrings for short obvious functions.  Add a pylint pragma to the function definition line.  That doesn't seem like too much to ask.

3. I don't see how removing a reminder to write docstrings can help us get better docstrings where we need them.

4. Code reviewers should be critiquing poor docstrings, and ensuring they are well written and helpful.

jzoldak:
@cpennington pointed out to me that we already have `docstring-min-length=5` which is what I was getting at anyhow!

PR #60, opened 2018-01-22
pin pylint-django-version

estute:
We previously pinned pylint-django from 0.7.2 to 1.0.0. However, 0.8 has created dependency issues in repos that consume edx-lint. Since 0.7.2 is the only release prior to 0.8, I am explicitly pinning it until we have another solution.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 pylint==1.7.1
 astroid==1.5.2
-pylint-django>=0.7.2,<1.0.0
+pylint-django==0.7.2
 pylint-celery==0.3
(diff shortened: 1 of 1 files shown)


sshishov:
Not very good approach as using `edx-lint` I cannot use latest `pylint` and `pylint-django`.

nedbat:
@sshishov We are working on ways to stay better up-to-date with third-party packages.  Using edx-lint gives us uniformity across our repos.  Before edx-lint, developers complained that they were being subjected to different quality rules in different repos, and it was confusing.

PR #64, opened 2018-10-27
Switch to pytest

nedbat:
Switching was easy.  Quieting the warnings was a journey through multiple levels of confusion, finally culminating in a blog post (linked from the tox.ini).

### DIFF
--- a/test/test_plugin_range_check.py
+++ b/test/test_plugin_range_check.py
@@ -1,18 +1,17 @@
 """Test range_check.py"""
 
 import astroid
-import ddt
-from pylint.testutils import Message
+from pylint.testutils import CheckerTestCase, Message
+import pytest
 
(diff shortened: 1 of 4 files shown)


test/test_plugin_range_check.py
```diff
@@ -1,18 +1,17 @@
...
+from .utils import get_module
 
 
-@ddt.ddt
-class RangeCheckerTest(PluginUnittestTestCase):
+class TestRangeCheckerTest(CheckerTestCase):
     CHECKER_CLASS = RangeChecker
 
-    @ddt.data("range", "xrange")
+    @pytest.mark.parametrize("range_name", ["range", "xrange"])
```
cpennington: You were able to get `parametrize` to work with a method? I had issues with that the other day... Maybe because I was deriving from a `unittest.TestCase`.

PR #65, opened 2018-10-28
Convert all the pylint plugin tests to new-style testing

nedbat:
This includes the conversion to pytest, which is in a [separate pull request](https://github.com/edx/edx-lint/pull/64).


nedbat:
@cpennington btw, let me know if you are tired of these :)

PR #66, opened 2018-10-29
Check setUpTestData for super calls.

nedbat:
Django has a setUpTestData method on its test cases that should also be checked for super() calls.  BTW, this is all I wanted to do, and it led to all of the recent edx-lint changes....!

### DIFF
--- a/test/plugins/test_super_check.py
+++ b/test/plugins/test_super_check.py
@@ -19,7 +19,7 @@ class TestUnitTestSetupSuperChecker(CheckerTestCase):
         "setUpClass",
         "tearDownClass",
     ])
-    def test_super_check(self, method):
+    def test_unittest_super_check(self, method):
         bad_node = astroid.extract_node("""
             import unittest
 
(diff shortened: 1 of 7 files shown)


test/plugins/test_super_check.py
```diff
@@ -45,7 +45,32 @@ def {method}(xyzzy):
...
             node=bad_node,
-            args=(method, 'TestCase'),
+            args=(method, 'unittest.case.TestCase'),
+        )
+        with self.assertAddsMessages(expected):
+            self.walk(module)
+
+    @pytest.mark.parametrize("method", [
+        "setUpTestData",
+    ])
```
jmbowman: Is this just parameterized in case more methods like this get added to the base class later?

tox.ini
```diff
@@ -4,6 +4,7 @@ envlist = py{27,35},coverage,pylint
 [testenv]
 deps =
     coverage
+    Django>=1.8,<1.9
```
jmbowman: This should probably be updated to use Django 1.11, since 1.8 isn't supported anymore.

test/plugins/test_super_check.py
nedbat: It's the same pattern as the code above.  I thought about doubly-parameterizing one method instead.  But it got a little complex specify the base class, and the real name of the base class, etc.

tox.ini
nedbat: Will do.

nedbat:
@jmbowman what say you?

PR #68, opened 2018-11-02
Don't build PyPI kits twice

nedbat:
Before this commit, Travis would try to upload the sdist distribution
for both 2.7 (twice), and 3.5.  That would fail because it already was
uploaded once, so the entire Travis build would be marked as a failure.


.travis.yml
```diff
@@ -1,10 +1,11 @@
...
 
-python:
-  - 2.7
-  - 3.5
 matrix:
  include:
+  - python: "2.7"
+    env: DISTRIBUTIONS="sdist bdist_wheel"
+  - python: "3.5"
+    env: DISTRIBUTIONS="bdist_wheel"
```
jmbowman: Is there a reason why this can't be a universal wheel that works with all Python versions?  I wasn't aware of any C extensions in this package.

nedbat: Good point! I've updated it to be a universal wheel.

.travis.yml
```diff
@@ -1,10 +1,10 @@
 language: python
 
-python:
-  - 2.7
-  - 3.5
 matrix:
  include:
+  - python: "2.7"
+    env: DISTRIBUTIONS="sdist bdist_wheel --universal"
```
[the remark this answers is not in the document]
nedbat: (and I like having the --universal switch next to the bdist_wheel command it affects.)

PR #70, opened 2018-11-02
Allow for hyphenated strings in the getattr check

georgebabey:
Our rest-api-client is based on slumber, which uses attributes on the api-client to build URL strings. URL paths often contain hyphenated strings so we allow getattr in this case to access these properties.


nedbat:
I really appreciate the pull request, but I beat you to it: https://github.com/edx/edx-lint/pull/69 :)

PR #74, opened 2018-11-18
Convert tests to integration tests

nedbat:
1. Create a helper, `run_pylint` that runs all of pylint for a test.  This has a few benefits:
    - Removes the coupling of our tests to internal details like `visit_call`.
    - Lets us test ITokenCheckers. The pylint docs have no examples of how to test those.
    - Lets us test all aspects of our checkers, including whether they can be disabled with pragmas.
1. Use tox to run on Python 2.7 and Python 3.6, and also on differing pylint versions, so we'll have better advance warning if the internals change out from under us.
1. Remove required_base_class checker: I forget why we wrote it, but nothing in the edx organization is using it.
1. Add a missing check to simplifiable-range.


nedbat:
@cpennington I hope this isn't too much to digest in one sitting.

PR #75, opened 2018-11-18
Check that human-readable format strings are Unicode

nedbat:
(This is into nedbat/integration-tests to keep the diff reasonable. I'll merge it to master once that PR is merged.)


pylintrc_tweaks
```diff
@@ -2,3 +2,6 @@
 [MASTER]
 ignore+ = ,input
 load-plugins = edx_lint.pylint
+
+[MESSAGES CONTROL]
+disable+ = ,unicode-format-string
```
nedbat: This might be controversial.

nedbat:
@cpennington A pragmatic check for the problem we stumbled on last week?

nedbat:
This introduces about 2800 unicode-format-string violations in edx-platform: https://github.com/edx/edx-platform/pull/19281

pylintrc_tweaks
cpennington: What does this particular flag do? Disable this check by default?

nedbat: This disables it for this repo's own pylint checks.  It's on by default for repos making use of edx-lint.

cpennington: Any particular reason to disable it here?

nedbat: I knew it would be controversial :)   No reason, other than this repo doesn't traffic in non-ASCII data.  I'll leave it enabled and fix the violations.

nedbat:
@cpennington Look ok?

PR #83, opened 2019-01-24
do not count docstrings as pylint violations

MatthewPiatetsky:


edx_lint/pylint/unicode_check.py
```diff
@@ -84,4 +97,12 @@ def process_tokens(self, tokens):
                     # Valid JSON isn't human-readable, leave it alone.
                     continue    # pragma: no cover   grr.. https://github.com/nedbat/coveragepy/issues/198
 
+                if re.match(r"^(def|class) [a-zA-Z0-9_]*\(.*", previous_line):
```
MatthewPiatetsky: although this may not always work, it may be good enough and may be simpler than other more correct solutions that would involve a lot more refactoring

test/plugins/test_unicode_check.py
```diff
@@ -11,14 +11,17 @@
...
-        "this is fine"
-        "this is not {}".format("fine")         #=A
-        "This is not %s" % ("fine",)            #=B
-        '{"json": "is fine"}'
-        u"Unicode strings are {}".format("fine")
-        b"byte strings are %s." % ("fine,")
-        "onewordis{}".format("fine")
-        """
+    def test():
+        \"docstring this is {fine}\"
```
nedbat: You don't need the backslashes here, do you?

edx_lint/pylint/unicode_check.py
```diff
@@ -58,7 +58,20 @@ def process_tokens(self, tokens):
             # If the whole module is __future__'d, then it's all fine.
             return
 
-        for tok_type, tok_text, start, _, _ in tokens:
+        previous_line = ""
+        current_line = ""
+        for index, tkn in enumerate(tokens):
+            tok_type, tok_text, start, _, _ = tkn
```
nedbat: This can be:
```
for index, (tok_type, tok_text, start, _, _) in enumerate(tokens):
```

test/plugins/test_unicode_check.py
```diff
@@ -11,14 +11,17 @@
...
+        pass
+    "this is fine"
+    "this is not {}".format("fine")         #=A
+    "This is not %s" % ("fine",)            #=B
+    '{"json": "is fine"}'
+    u"Unicode strings are {}".format("fine")
+    b"byte strings are %s." % ("fine,")
+    "onewordis{}".format("fine")
```
nedbat: Let's add a class docstring, and a method docstring to this test data also.

edx_lint/pylint/unicode_check.py
nedbat: A small tweak to the regex:
```
re.match(r"^\s*(def|class) [a-zA-Z0-9_]+\(", previous_line)
```

edx_lint/pylint/unicode_check.py
```diff
@@ -84,4 +97,12 @@ def process_tokens(self, tokens):
                     # Valid JSON isn't human-readable, leave it alone.
                     continue    # pragma: no cover   grr.. https://github.com/nedbat/coveragepy/issues/198
 
+                if re.match(r"^(def|class) [a-zA-Z0-9_]*\(.*", previous_line):
+                    # Previous line probably contains a function or class definition
+                    # so this string is probably a docstring
+                    continue
+                if index == 0:
+                    # If this string is the first line of the file, it is likely a docstring
```
nedbat: "If this string is the first token of the file, it is definitely a docstring"

But the docstring can appear after a comment on the first line, so it isn't necessarily the first token.

edx_lint/pylint/unicode_check.py
nedbat: Also, let's move this re.match check before the json.loads attempt.  No point trying to json-parse all the docstrings in the code.

MatthewPiatetsky: good point, i'll move it before.  
I think it may have worked without the s due to indents being another token but i'll add it just in case. we still want the .* at the end though since this is match not search i think

nedbat: `match` only needs to match the start of the string (and in fact, `^` is redundant for match), so the trailing `.*` can be removed.

edx_lint/pylint/unicode_check.py
MatthewPiatetsky: good point, added code to check for comments before the file docstring

edx_lint/pylint/unicode_check.py
MatthewPiatetsky: ah ok good to know, thanks

edx_lint/pylint/unicode_check.py
```diff
@@ -58,7 +58,22 @@ def process_tokens(self, tokens):
             # If the whole module is __future__'d, then it's all fine.
             return
 
-        for tok_type, tok_text, start, _, _ in tokens:
+        previous_line = ""
+        current_line = ""
+        first_non_comment_line = 0
+        for index, (tok_type, tok_text, start, _, _) in enumerate(tokens):
+            # Use new line characters to build previous line
+            if first_non_comment_line == 0 and tok_type != 53: # 53 is token.COMMENT
```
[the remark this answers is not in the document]
nedbat: Pylint has an improved tokenizer that also provides comments: https://kite.com/python/docs/pylint.utils.tokenize

edx_lint/pylint/unicode_check.py
```diff
@@ -58,7 +58,22 @@ def process_tokens(self, tokens):
...
+                first_non_comment_line = index
+
+            if tok_type == token.NEWLINE or tok_type == 54:  # 54 is token.NL
+                previous_line = current_line
+                current_line = ""
+            else:
+                current_line += tok_text
+                if tok_text in ['def', 'class']:
+                    current_line += " "
```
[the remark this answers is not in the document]
nedbat: We only need `current_line` for checking if a previous line is a definition.  It could be simpler to just always add a space between tokens, and then update the regex to match the result.  Or: don't keep the previous line text at all, just remember if there was a keyword "def" or "class" token in the previous line.

edx_lint/pylint/unicode_check.py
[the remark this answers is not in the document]
nedbat: The variable is called `first_non_comment_line`, but isn't a line or line number, it's a token number.  The first comment will be token 0, so the initialization to 0 seems wrong.  Use `None` instead.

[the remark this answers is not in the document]
nedbat: `first_non_comment_line` isn't a line or line number, it's a token index.  Tokens start at 0, so the init value should be `None`, not 0.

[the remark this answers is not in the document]
MatthewPiatetsky: i believe token.COMMENT is only defined for python 3 and was gave me an error when I tried to use it in the code.  
however, tok_type was 53 and token.tok_name[53] was 'COMMENT'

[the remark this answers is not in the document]
MatthewPiatetsky: hmm..well I'm iterating over tokens, not lines, but comments at the beginning of the file are treated as a single token.  
i guess i could call this variable first_non_comment_token, but I think the intent might be clearer if I call it first_non_comment_line, since if I have a string on the first non-comment line it is a docstring?  
I'm open to renaming it though

edx_lint/pylint/unicode_check.py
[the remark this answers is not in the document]
MatthewPiatetsky: i was just adding the space after those tokens for my regex so it wouldn't get confused by functions or variables that contain those strings as substring.  
true, i can update the code to only check for those tokens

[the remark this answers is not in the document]
MatthewPiatetsky: updated the code

PR #84, opened 2019-01-29
Check that we can disable unicode-format-string

nedbat:


nedbat:
@robrap I was a good boy.

nedbat:
Yes, I checked removing the pragma: the test fails.  I fixed the text of the string also.

PR #85, opened 2019-01-31
WIP: Feature toggle linting

robrap:
Still needs some TLC.

### DIFF
--- a/edx_lint/pylint/feature_toggle_check.py
+++ b/edx_lint/pylint/feature_toggle_check.py
@@ -1,0 +1,180 @@
+"""Pylint plugin: checks that feature toggles are properly annotated."""
+
+import re
+
+from pylint.checkers import BaseChecker, utils
+from pylint.interfaces import IAstroidChecker
+
+from .common import BASE_ID, check_visitors
+
+
+def register_checkers(linter):
+    """Register checkers."""
+    linter.register_checker(FeatureToggleChecker(linter))
+
+class AnnotationLines(object):
+    """
+    StringLines provides utility methods to work with a string in terms of
+    lines.  As an example, it can convert an index into a line number or column
+    number (i.e. index into the line).
+    """
+
+    # Regex searches for annotations like: [    # .. ]
+    _ANNOTATION_REGEX = re.compile(r'[\s\S]*#[\s\S]*?\.\.[\s\S]*?feature_toggle')
+
+    def __init__(self, module_node):
+        """
+        Arguments:
+            module_node: The visited module node.
+        """
+        module_as_binary = module_node.stream().read()
+        self._string = module_as_binary.decode("iso8859-1")
+
+        self._line_start_indexes = self._process_line_breaks(self._string)
+        # this is an exclusive index used in the case that the template doesn't
+        # end with a new line
+        self.eof_index = len(self._string)
+
+    def is_line_annotated(self, line_number):
+        """
+        Checks if the provided line number is annotated.
+        """
+        if line_number < 1 or self.line_count() < line_number:
+            return False
+
+        return bool(
+            self._ANNOTATION_REGEX.match(self._line_number_to_line(line_number))
+        )
+
+    def line_count(self):
+        """
+        Gets the number of lines in the string.
+        """
+        return len(self._line_start_indexes)
+
+    def _process_line_breaks(self, string):
+        """
+        Creates a list, where each entry represents the index into the string
+        where the next line break was found.
+
+        Arguments:
+            string: The string in which to find line breaks.
+
+        Returns:
+             A list of indices into the string at which each line begins.
+
+        """
+        line_start_indexes = [0]
+        index = 0
+        while True:
+            index = string.find('\n', index)
+            if index < 0:
+                break
+            index += 1
+            line_start_indexes.append(index)
+        return line_start_indexes
+
+    def _line_number_to_line(self, line_number):
+        """
+        Gets the line of text designated by the provided line number.
+
+        Arguments:
+            line_number: The line number of the line we want to find.
+
+        Returns:
+            The line of text designated by the provided line number.
+
+        """
+        start_index = self._line_start_indexes[line_number - 1]
+        if len(self._line_start_indexes) == line_number:
+            line = self._string[start_index:]
+        else:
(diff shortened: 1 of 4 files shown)


edx_lint/pylint/feature_toggle_check.py
```diff
@@ -0,0 +1,186 @@
...
+    def line_count(self):
+        """
+        Gets the number of lines in the string.
+        """
+        return len(self._line_start_indexes)
+
+    def _process_line_breaks(self, string):
+        """
+        Creates a list, where each entry represents the index into the string
+        where the next line break was found.
```
nedbat: This sounds like a complex thing that could be done more easily by just keeping a list whose elements were the lines of source text.

edx_lint/pylint/feature_toggle_check.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+
+    __implements__ = (IAstroidChecker,)
+
+    name = 'feature-toggle-annotation-checker'
+
+    WAFFLE_NOT_ANNOTATED_MESSAGE_ID = 'waffle-missing-toggle-annotation'
+    ILLEGAL_WAFFLE_MESSAGE_ID = 'illegal-waffle-usage'
+
+    _CHECK_CAPITAL_REGEX = re.compile(r'[A-Z]')
```
nedbat: Stylistically, I prefer to avoid re.match and always use re.search:  `r'^[A-Z]'`

edx_lint/pylint/feature_toggle_check.py
```diff
@@ -0,0 +1,186 @@
...
+        if not hasattr(node.func, 'name'):
+            return
+
+        # Looking for class instantiation, so should start with a capital letter
+        starts_with_capital = self._CHECK_CAPITAL_REGEX.match(node.func.name)
+        if not starts_with_capital:
+            return
+
+        # Search for toggle classes that require an annotation
+        if not any(node.func.name.endswith(toggle_class) for toggle_class in self._WAFFLE_TOGGLE_CLASSES):
```
nedbat: Fun fact: endswith can take a tuple of possibilities, and check them all.

edx_lint/pylint/feature_toggle_check.py
```diff
@@ -0,0 +1,186 @@
...
+class FeatureToggleChecker(BaseChecker):
+    """
+    Checks that feature toggles are properly annotated.
+    """
+
+    __implements__ = (IAstroidChecker,)
+
+    name = 'feature-toggle-annotation-checker'
+
+    WAFFLE_NOT_ANNOTATED_MESSAGE_ID = 'waffle-missing-toggle-annotation'
```
nedbat: As this checker grows to check more things, it might be better for the message id to be "feature-toggle-missing-annotation", and use one message id for all the kinds of feature toggles.  (Also, it's kind of long, how about "feature-toggle-needs-doc"?)

nedbat: The text of the message could include the type of the toggle, just that one message id might make it easier to search for, etc.

edx_lint/pylint/feature_toggle_check.py
```diff
@@ -0,0 +1,186 @@
...
+            )
+
+    def check_illegal_waffle_usage(self, node):
+        """
+        Check Call node for illegal waffle calls.
+        """
+        if not hasattr(node.func, 'name'):
+            return
+
+        if node.func.name in self._ILLEGAL_WAFFLE_FUNCTIONS:
```
nedbat: I was ok with just checking the name for "WaffleFlag", but, "flag_is_active" seems like a higher chance for false positives.  How do you feel about checking for "waffle.flag_is_active"?

edx_lint/pylint/feature_toggle_check.py
```diff
@@ -0,0 +1,186 @@
...
+
+class AnnotationLines(object):
+    """
+    StringLines provides utility methods to work with a string in terms of
+    lines.  As an example, it can convert an index into a line number or column
+    number (i.e. index into the line).
+    """
+
+    # Regex searches for annotations like: [    # .. ]
+    _ANNOTATION_REGEX = re.compile(r'[\s\S]*#[\s\S]*?\.\.[\s\S]*?feature_toggle')
```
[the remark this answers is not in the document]
robrap: Copy/pasted from a bad example and I didn't know what `\S` was and forgot.  Can correct.

edx_lint/pylint/feature_toggle_check.py
```diff
@@ -0,0 +1,186 @@
...
+from .common import BASE_ID, check_visitors
+
+
+def register_checkers(linter):
+    """Register checkers."""
+    linter.register_checker(FeatureToggleChecker(linter))
+
+class AnnotationLines(object):
+    """
+    StringLines provides utility methods to work with a string in terms of
```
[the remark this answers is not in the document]
robrap: I had written StringLines for the XSS linter and just reused it and added some annotation helpers.

edx_lint/pylint/feature_toggle_check.py
robrap: As noted, I just grabbed old code that was doing this from the XSS linter.  I agree something simpler would work.

edx_lint/pylint/feature_toggle_check.py
robrap: Ok.

edx_lint/pylint/feature_toggle_check.py
robrap: 👍

edx_lint/pylint/feature_toggle_check.py
robrap: I had it that way at first, and changed it so that documentation could be more clear because it wouldn't have to say "If this, then do this, if this, then do this, ...".  I don't feel strongly about it though, so happy to use one id.
FYI: Reusing the same linter ID does break things I learned.  At least, my tests started failing and I didn't know why.

robrap:
@jmbowman: FYI: I'm closing this because I don't plan to continue with this.  Feel free to reopen or restart the branch when and if your team picks up this work again.

PR #88, opened 2019-02-13
Add a check for unsafe yaml.load() calls.

iloveagent57:
https://nvd.nist.gov/vuln/detail/CVE-2017-18342

> In PyYAML before 4.1, the yaml.load() API could execute arbitrary code. In other words, yaml.safe_load is not used.

There's not a stable release beyond 3.13 yet (which is the version of `pyyaml` we're using), so let's go with this in the meantime?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ build/
 dist/
 
 .eggs/
+
+#pyenv
+.python-version
--- a/edx_lint/pylint/yaml_load_check.py
+++ b/edx_lint/pylint/yaml_load_check.py
@@ -1,0 +1,51 @@
+"""
+Checks for unsafe ``yaml.load()`` calls.
+The PyYAML library has not yet released a version
+that addresses the vulnerability (listed below) in the commonly
+called ``load()`` function.
+
+Vulnerability: https://nvd.nist.gov/vuln/detail/CVE-2017-18342
+PyYAML release plan: https://github.com/yaml/pyyaml/issues/193
+
+When a new release is published, we may be able to remove this check.
+"""
+
+from pylint.checkers import BaseChecker, utils
+from pylint.interfaces import IAstroidChecker
+
+from .common import BASE_ID, check_visitors
+
+
+def register_checkers(linter):
+    """Register checkers."""
+    linter.register_checker(YamlLoadChecker(linter))
+
+
+@check_visitors
+class YamlLoadChecker(BaseChecker):
+    """
+    Checks for unsafe ``yaml.load()`` calls.
+    """
+
+    __implements__ = (IAstroidChecker,)
+
+    name = 'yaml-load-checker'
+
+    MESSAGE_ID = 'unsafe-yaml-load'
+
+    UNSAFE_CALLS = {'yaml.load', 'yaml.load_all'}
+
+    msgs = {
+        'C{}57'.format(BASE_ID): (
+            u'yaml.load%s() call is unsafe, use yaml.safe_load%s()',
+            MESSAGE_ID,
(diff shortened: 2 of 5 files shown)


iloveagent57:
Whoops, just realized there were unit tests for this repo :)

nedbat:
Let me know if it isn't clear how to write new tests. It's a little idiosyncratic.

iloveagent57:
@nedbat thanks, I think it's ready for review now.

test/plugins/test_yaml_load_check.py
```diff
@@ -0,0 +1,30 @@
...
+
+    yaml.load('foo.bar')
+    yaml.load()
+    yaml.load('foo.bar', kwargs=**{'a': 'b'})
+    """
+    messages = run_pylint(source, MSG_IDS)
+    expected_msg = 'yaml.load() call is unsafe (unsafe-yaml-load)'
+
+    for actual_msg in messages:
+        assert expected_msg in actual_msg
```
jmbowman: No assertion regarding the number of messages?

iloveagent57: Hah, that would indeed be helpful :)

.gitignore
```diff
@@ -18,3 +18,6 @@ build/
 dist/
 
 .eggs/
+
+#pyenv
+.python-version
```
nedbat: I haven't seen this before: what is it?

edx_lint/pylint/yaml_load_check.py
```diff
@@ -0,0 +1,49 @@
...
+    Checks for unsafe ``yaml.load()`` calls.
+    """
+
+    __implements__ = (IAstroidChecker,)
+
+    name = 'yaml-load-checker'
+
+    MESSAGE_ID = 'unsafe-yaml-load'
+
+    UNSAFE_CALL = 'yaml.load'
```
nedbat: Can we also check for the lesser-used "load_all"?  BTW, there's also "safe_dump" and "safe_dump_all", though I'm less clear on what the dangers are with dump?
```
>>> [n for n in dir(yaml) if "safe" in n]
['safe_dump', 'safe_dump_all', 'safe_load', 'safe_load_all']
```

test/plugins/test_yaml_load_check.py
nedbat: It's also good to tie the messages to specific lines, and to add some pragmas to the test to be sure they work properly for this checker.

.gitignore
iloveagent57: It's generated by pyenv: https://github.com/pyenv/pyenv . Many of us use it for virtualenv management, particularly in repos in which we develop outside of a devstack context.

edx_lint/pylint/yaml_load_check.py
iloveagent57: I think the "safe_*" functions are meant to be safe to use.  The goal of this checker is to warn against `load()` and have the developer use `safe_load()` instead.  At least, that was my reading of the PyYAML release plan and the CVE.

iloveagent57: We can check for `load_all`, though - good call on that.

.gitignore
nedbat: I see.  You might like to create a global gitignore file to fix this across all your repos also: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

iloveagent57: Wow, good tip, thanks for the knowledge!

iloveagent57:
@nedbat This is ready for review again when you have a chance, thanks!

PR #90, opened 2019-02-16
Adapt to a pylint 2.0 name change

nedbat:


nedbat:
@eric-s-s This fixes the name error
@cpennington @jmbowman Opinions?

edx_lint/pylint/common.py
```diff
@@ -1,11 +1,18 @@
...
 
-from astroid.nodes import ALL_NODE_CLASSES
+import astroid
+import astroid.nodes
+
+# Pylint changed the name of the uninferable indicator in 2.0
+try:
+    Uninferable = astroid.Uninferable
+except AttributeError:
+    Uninferable = astroid.YES
```
jmbowman: It looks like `astroid.Uninferable` [was introduced](https://github.com/PyCQA/astroid/commit/3ca0c7580187913bec11d0042a36906a2af36c91) back in astroid 1.5.0 over 3 years ago; they just [removed the alias](https://github.com/PyCQA/astroid/commit/36fa294e9b84885d95a03b1fbc07da63e6d6a79b) for the original name recently.  Given that we've been pinning edx-lint to astroid 1.5.3 anyway, I think it's ok to just use the new name without the fallback.

nedbat: Yoinks! Good catch.  Fixed.

PR #91, opened 2019-02-21
Disable open-builtin pylint violation.

doctoryes:
I don't see a good reason for this violation. It seems that `open()` is still the best practice for opening a file in Python3. If I'm incorrect, please educate me!

### DIFF
--- a/edx_lint/files/pylintrc
+++ b/edx_lint/files/pylintrc
@@ -307,6 +307,7 @@ disable=
     old-ne-operator,
     old-octal-literal,
     old-raise-syntax,
+    open-builtin,
     parameter-unpacking,
     print-statement,
     raising-string,


doctoryes:
I just educated myself a bit more - it seems I should be doing this:
`from io import open`
I assume you agree that's preferable to disabling this pylint violation?

jmbowman:
I think the issue here is that the behavior of `open()` [changed subtly](https://docs.python.org/3.0/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit) in Python 3, and `io.open()` was added in 2.6 using the new behavior.  So the pylint check is there (in Python 2 only) to warn you that the code might break in corner cases after the upgrade.  After the upgrade, `open()` and `io.open()` are equivalent so it doesn't really matter which one you use.

nedbat:
What is the change in behavior.  From that linked page: "The builtin open() function is now an alias for io.open()"

jmbowman:
> Files opened as text files (still the default mode for open()) always use an encoding to map between strings (in memory) and bytes (on disk). Binary files (opened with a b in the mode argument) always use bytes in memory. This means that if a file is opened using an incorrect mode or encoding, I/O will likely fail loudly, instead of silently producing incorrect data. It also means that even Unix users will have to specify the correct mode (text or binary) when opening a file. There is a platform-dependent default encoding, which on Unixy platforms can be set with the LANG environment variable (and sometimes also with some other platform-specific locale-related environment variables). In many cases, but not all, the system default is UTF-8; you should never count on this default. Any application reading or writing more than pure ASCII text should probably have a way to override the encoding.

PR #96, opened 2019-02-25
Don't test py37-pylint1x

nedbat:
Python 3.7 fails when running Pylint 1.x.  We don't need it to work, so don't test it.


nedbat:
@cclauss ??

PR #101, opened 2019-05-24
Feature toggle annotation linter

michaelyoungstrom:
Working off previous hackathon work by Robert: https://github.com/edx/edx-lint/pull/85. 
Documentation: https://openedx.atlassian.net/wiki/spaces/HACK/pages/940704078/Documenting+feature+toggles

Adds a linter for feature toggle annotations. The expectation is to have some documentation along the lines of [this example](https://github.com/edx/edx-platform/blob/d5ee03e81d06bfe067d51392d5827763992a3bba/lms/djangoapps/experiments/utils.py#L35-L50) for each feature toggle. There is at least one instance where we document a toggle with `documented_elsewhere`: [link to example](https://github.com/edx/edx-platform/blob/b9c7c10ca7416eeb8cff2181458b188378549d6b/cms/envs/common.py#L184), so I also accept that as valid documentation. 

To check this we simply look at the line above a toggle to make sure it has something that looks like 
`# .. toggle`
or
`# .. documented_elsewhere`

Toggle types to lint include WaffleFlag, WaffleSwitch, ConfigurationModels, and Django Feature Flags.

For Django Feature Flags, I'm starting by just explicitly looking at the`FEATURES` dict creation, like [this example](https://github.com/edx/edx-platform/blob/5eac6aa0491adf5f3ab5ef1a56d9657cf6aa3715/lms/envs/common.py#L81) . We can look into expanding this if desired, as there are definitely flags that get created after the fact, like [this example](https://github.com/edx/edx-platform/blob/5eac6aa0491adf5f3ab5ef1a56d9657cf6aa3715/lms/envs/bok_choy.py#L108). 

Unfortunately, for the features dict itself, I think we may be limited by the possible `visit_` methods in astroid, as the only applicable method I can find it `visit_dict`. This means the method only gets called once (when a dict gets created) so a pylint suppression can only be applied to all or none of the dict. It would be nice to have more granularity here if possible but I haven't found anything to help with this yet.

### DIFF
--- a/edx_lint/pylint/unicode_check.py
+++ b/edx_lint/pylint/unicode_check.py
@@ -21,10 +21,9 @@ def register_checkers(linter):
 @check_visitors
 class UnicodeFormatStringChecker(BaseTokenChecker):
     """
-    XXXXXX
+    Checks that strings are unicode.
 
-    The message id is xxxxxxx.
-
+    Message ID is: unicode-format-string
     """
     # ITokenChecker gets us process_tokens support.
     # IRawChecker gets us process_module support.
--- a/edx_lint/pylint/feature_toggle_check.py
+++ b/edx_lint/pylint/feature_toggle_check.py
@@ -1,0 +1,217 @@
+"""
+Pylint plugin: checks that feature toggles are properly annotated.
+"""
+
+from __future__ import absolute_import
+import re
+
+from pylint.checkers import BaseChecker, utils
+from pylint.interfaces import IAstroidChecker
+
+from .common import BASE_ID, check_visitors
+
+
+def register_checkers(linter):
+    """
+    Register checkers.
+    """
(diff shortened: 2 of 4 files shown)


edx_lint/pylint/unicode_check.py
```diff
@@ -21,10 +21,7 @@ def register_checkers(linter):
 @check_visitors
 class UnicodeFormatStringChecker(BaseTokenChecker):
     """
-    XXXXXX
-
-    The message id is xxxxxxx.
-
+    Checks that strings are unicode.
```
jmbowman: Should probably also state the actual ID for the message, as suggested by the prior placeholder template.

michaelyoungstrom:
@estute No time like the present!

jmbowman:
👍

PR #103, opened 2019-09-30
Switch over to standard setuptools setup.cfg usage

cpennington:


setup.cfg
```diff
@@ -16,16 +17,22 @@ classifiers =
...
-    edx_lint.pylint
+[options]
+packages = find:
+install_requires =
+    pylint==1.7.6
+    pylint-django==0.7.2
+    pylint-celery==0.3
+    six>=1.10.0,<2.0.0
+    click>=6.0
+    click-log==0.3.2
```
nedbat: How does this list tap into the pip-compile workflow? Or it doesn't?

cpennington: It does for `dev.in`/`dev.txt`. However, my expectation is that since edx-lint is intended to standardize pylint versions, that `setup.py` should pin particular pylint versions.

nedbat: It feels like we need a note someplace that these lines are outside the "make upgrade" flow.  How will people know to update these pins?

cpennington: I just added a message at the end of `make upgrade`

cpennington:
Oops. I thought I had squashed everything...

PR #104, opened 2019-09-30
Upgrade pylint

cpennington:


setup.cfg
```diff
@@ -20,9 +20,11 @@ classifiers =
 [options]
 packages = find:
 install_requires =
-    pylint==1.7.6
-    pylint-django==0.7.2
-    pylint-celery==0.3
+    pylint==1.9.5;python_version<"3.0"  # The last pylint version to support python 2
+    pylint==2.4.2;python_version>="3.0"
```
nedbat: Do we run the risk of warnings vacillating as the same repo is linted sometimes with python 2, sometimes with python 3?

setup.cfg
```diff
@@ -20,9 +20,11 @@ classifiers =
...
+    pylint-django==0.11.1;python_version<"3.0"  # The last pylint-django to support python 2
+    pylint==2.0.11;python_version>="3.0"
```
nedbat: This line is a duplicate of line 24, but with a different version?

setup.cfg
```diff
@@ -20,9 +20,11 @@ classifiers =
...
+    pylint-celery
```
nedbat: We don't want to pin this package?

setup.cfg
cpennington: oops. supposed to be `pylint-django`

setup.cfg
cpennington: Maybe? But I think our linting jobs should just be pinned to either lint w/ 2, or lint w/ 3, but not both.

PR #120, opened 2020-07-16
Add docs for getting latest pylintrc

robrap:


edx_lint/cmd/write.py
```diff
@@ -23,6 +23,9 @@
 # you want to make a local change that applies only to this repo, or whether
 # you want to make a central change that applies to all repos using edx-lint.
 #
+# Note: If your pylintrc file is simply out-of-date relative to the latest
+# pylintrc rules in edx-lint, follow the steps for a "LOCAL CHANGE".
+#
```
nedbat, line 28: You might want to add, "install the latest edx-lint, then LOCAL CHANGE"

robrap: FYI: I updated, and also applied it to this repo.

PR #121, opened 2020-08-18
Adding common constraint file for edx repos usage.

awais786:
https://openedx.atlassian.net/browse/BOM-1867

Purpose of this file is to define a common constraint at one place and use it in all edx repos. e.g right now all edx repos using `django<2.3` but we have to mentioned it in all repos local constraint.txt file. But now you can just define the constraint in this file and call the path in ur desired edx repo constraint.txt like this 

`-c https://raw.githubusercontent.com/edx/edx-lint/84f463efed7f287a60eeb44cffa4884342066b2c/edx_lint/files/common_constraints.txt`

If you don't need the constraint version you can overwrite in local constraint file.

### DIFF
--- a/edx_lint/files/common_constraints.txt
+++ b/edx_lint/files/common_constraints.txt
@@ -1,0 +1,3 @@
+# common constraints file across edx usage
+# using LTS django version
+Django<2.3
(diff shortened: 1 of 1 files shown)


nedbat:
I'm not sure how draft-y this PR is, so apologies if I am jumping in too soon.  Putting this file in edx_lint/files makes sense to me.

awais786:
created test PR for credentials. The link needs to be updated there after merging this PR.
https://github.com/edx/credentials/pull/812/files#diff-294960731653f2fdb7b827504766ff5dR12

nedbat:
Thanks! Once the mechanisms are all worked out, let's put a big comment at the top of this file explaining how it is used.

PR #125, opened 2020-09-04
Revert "pytest-xdist latest version breaking tests. It needs to pin"

awais786:
Reverts edx/edx-lint#124


mzulqarnain1 (approved):

awais786:
This is not widely in use so its not a good point to add it globally.

PR #126, opened 2020-09-08
Upgrade pylint

cpennington:


cpennington:
No, I'm not sure what the level of changes is going to be. I started doing this because ecommerce was having issues w/ a checker that should be resolved in a later version of `pylint-django`, and then after that I realized that we probably need to update the list of checkers, since we're a) in python 3, and b) in a newer version of pylint that probably has useful checks. But, I think we're just going to need a trial run of upgrading edx-lint.

cpennington:
As far as rollout plan: I'm going to be using it as the basis for adding edx-lint to ecommerce, first. I expect edx-platform to pick up the new version w/ the automatic upgrade script, and we'll have to see if that bumps the numbers up too badly.

Slightly longer term, I want to use the repo health dashboard to make sure that we're rolling out edx-lint files across the org (and keeping up w/ version updates).

feanil (approved):

edx_lint/cmd/write.py
```diff
@@ -82,6 +83,8 @@
 # SERIOUSLY.
 #
 # ------------------------------
+# VERSION: {version}
```
nedbat, line 86: nit: this is the version of edx-lint that produced the file, not the version of the file.  The written files in each repo can vary from each other because of local tweaks files.

cpennington: That's intentional, but I'll make it clearer in the tag. I want to have the version that generated the file so that we can validate that all repos are using the latest generated settings.

PR #132, opened 2020-11-30
Added GitHub workflows for CI and PyPi publish

mraarif:
Relevant JIRA issue here: https://openedx.atlassian.net/browse/BOM-2107


.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
+name: Python CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
nedbat, line 7: I think it's better to omit the branch name here, so that a pull request will always trigger actions, regardless of what branch it is destined for.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+
+
+jobs:
+  run_tests:
+    name: tests
+    runs-on: ${{ matrix.os }}
```
nedbat, line 13: Do we need this indirection if the matrix only has one OS?

.github/workflows/ci.yml
nedbat, line 7:
```suggestion
  pull_request:
```

.github/workflows/ci.yml
jmbowman: We started doing it this way because:

1. It shows the OS version in use in the description for each check, since it's a matrix parameter.  This is handy for sanity checking that we didn't miss updating any tests during the last OS upgrade.
1. It simplifies the changes needed the next time we do need to test 2 different OS versions during an upgrade cycle.

mraarif:
@cpennington @jmbowman I've noticed that `edx-lint` was updated to use `pylint 2.6` (https://github.com/edx/edx-lint/pull/126) and tox/travis config was never updated to install `pylint 2.6` (https://travis-ci.org/github/edx/edx-lint/jobs/726010462), if we install `pylint 2.6` there are quite a few tests that fail
as per my understanding `pylint24` tox environment tests against `pylint 2.4` and `pylint` tests against the latest `pylint` 
`pylint 2.6` only seems to be running the linting on `edx-lint` itself, is this intentional?

jmbowman:
It looks like we've been testing pylint 2.4 with Python 3.8 and pylint 2.6 with Python 3.5 (see https://travis-ci.com/github/edx/edx-lint/jobs/455327411 for the latter).  That probably isn't really what we wanted to do, but at least those two combinations seem to be passing.

mraarif:
@jmbowman the build here: https://travis-ci.com/github/edx/edx-lint/jobs/455327411 is running pylint on edx-lint itself, shouldn't it be running unit tests with pylint 2.6?

jmbowman:
It does:

```,pylint==2.6.0,```

It's just creating the virtualenv in a directory with a confusing name:

```envdir = {toxworkdir}/py35-pylint24```

There's definitely some cleanup to do here, but I think at least one of the workers is running pylint 2.6.0.  You could add `pylint --version` to the tox commands if you want to confirm.

**Review by mzulqarnain1 — APPROVED** [review]

**Comment by mraarif** [comment]
@jmbowman current Github workflow matrix setup should be Okay to merge?

File: .github/workflows/pypi-publish.yml [hunk 3]
```diff
@@ -0,0 +1,30 @@
...
+      - name: Install pip
+        run: pip install -r requirements/pip.txt
+
+      - name: Build package
+        run: python setup.py sdist bdist_wheel
+
+      - name: Publish to PyPi
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          user: edx
```
jmbowman, line 29: Shouldn't this be `__token__`?

jmbowman (approved):

.github/workflows/pypi-publish.yml
mraarif: yes, forgot this one

PR #140, opened 2021-01-28
[BD-21] Report generic code annotation errors

regisb:
Previously, it was required to run `code_annotations --lint` in order to verify
code annotations. The errors then had to be collected manually, for instance
with a paver task. This made it intractable to generalize code annotation
checking to multiple IDAs. With this new checker, code annotation errors are
sent to pylint, which makes it easy for everyone to quickly detect annotation
errors. To achieve this, we make use of the new code annotations API created in
v1.1.0.

This PR depends on https://github.com/edx/code-annotations/pull/61 (:heavy_check_mark: merged)

cc @robrap

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,18 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[4.0.0] - 2021-01-28
+~~~~~~~~~~~~~~~~~~~~
+
+* BREAKING CHANGE: modify the numerical ID of annotation checks
+* BREAKING CHANGES:
+
+  * modify the numerical ID of annotation checks
+  * though technically not a breaking change, the new annotation checks may break your build if there are pre-existing
+    violations.
+
+* Add ``CodeAnnotationChecker`` to run generic checks on annotations
+
 [3.0.2] - 2021-01-26
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/edx_lint/pylint/annotations_check.py
+++ b/edx_lint/pylint/annotations_check.py
@@ -7,6 +7,7 @@ import re
 
 import pkg_resources
 
+from code_annotations import annotation_errors
 from code_annotations.base import AnnotationConfig
 from code_annotations.find_static import StaticSearch
 from pylint.checkers import BaseChecker, utils
@@ -20,8 +21,23 @@ def register_checkers(linter):
     Register checkers.
     """
     linter.register_checker(FeatureToggleChecker(linter))
+    linter.register_checker(CodeAnnotationChecker(linter))
     linter.register_checker(FeatureToggleAnnotationChecker(linter))
     linter.register_checker(SettingAnnotationChecker(linter))
+
+
+def check_all_messages(msgs):
+    """
+    Decorator to automatically assign all messages from a class to the list of messages handled by a checker method.
+
+    Inspired by pylint.checkers.util.check_messages
+    """
+
+    def store_messages(func):
+        func.checks_msgs = [message[1] for message in msgs]
+        return func
+
+    return store_messages
 
 
 class AnnotationLines:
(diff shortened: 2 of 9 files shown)


edx-webhook:
Thanks for the pull request, @regisb! I've created [BLENDED-746](https://openedx.atlassian.net/browse/BLENDED-746) to keep track of it in Jira.   More details are on the [BD-21](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1596358943/BD-21+Toggles+Settings+Documentation) project page.


When this pull request is ready, tag your edX technical lead.

edx_lint/pylint/annotations_check.py
```diff
@@ -233,85 +249,126 @@ class AnnotationBaseChecker(BaseChecker):
...
         raise NotImplementedError
 
 
+class CodeAnnotationChecker(AnnotationBaseChecker):
+    """
+    Run generic code annotation checks.
+
+    This makes use of code-annotations `check_results()` method. Elements from
+    `search.annotation_errors` are then parsed and exposed to pylint. Note that modifying the creation order of the
+    error types in code-annotations will lead to a modification of the numerical IDs of the errors here.
```
robrap, line 290: Can we add something like: ", so we will ensure that doesn't happen." or ", so we will avoid that.".

edx_lint/pylint/annotations_check.py
```diff
@@ -233,85 +249,126 @@ class AnnotationBaseChecker(BaseChecker):
...
+
+    When creating a new annotation configuration, its filename should be added to
+    CodeAnnotationChecker.CONFIG_FILENAMES.
```
robrap, line 293: Do these files need to live in code-annotations contrib/config?  Should we mention that, potentially with a link to https://github.com/edx/code-annotations/tree/master/code_annotations/contrib/config?

edx_lint/pylint/annotations_check.py
```diff
@@ -233,85 +249,126 @@ class AnnotationBaseChecker(BaseChecker):
...
+    """
+    CONFIG_FILENAMES = ["feature_toggle_annotations.yaml", "setting_annotations.yaml"]
+    __implements__ = (IAstroidChecker,)
+    name = "code-annotations"
+    msgs = {
+        ("E%d%d" % (BASE_ID, index + 50)): (
```
robrap, line 299: 1. I'm not clear why this couldn't have defined ids ending in `7x`, and left all the ids below as-is?
2. May be silly, but in the potential comment and unit test detailed in [this code-annotations PR comment](https://github.com/edx/code-annotations/pull/61/files#r566527416), should we also mention the max of 10? Also, since we already have 6, should we leave more room? Note: if we did make this `7x`, then leaving more room would be somehow letting people know to jump to 9x?

CHANGELOG.rst
```diff
@@ -13,6 +13,12 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[4.0.0] - 2021-01-28
+~~~~~~~~~~~~~~~~~~~~
+
+* Add CodeAnnotationChecker to run generic checks on annotations
+* BREAKING CHANGE: modify the numerical ID of annotation checks
```
robrap, line 20: Should the new checks also be mentioned under the breaking change?  I wonder if any new linting checks should be considered a breaking change, because if someone has a threshold of zero and new violations without amnesty, it will break their build. Food for thought.

robrap (commented):
Thanks again.  I'm looking forward to landing this.

Am I being overly sensitive to violation ID changes, and potential future changes? I feel like there may be some config files that only allow us to use the ids, which is where my concern lies.

@jmbowman: Are there any places where we must use the pylint violation IDs, or am I just remembering pydocstyle like here: https://github.com/edx/edx-cookiecutters/blob/master/python-template/%7B%7Bcookiecutter.placeholder_repo_name%7D%7D/tox.ini#L30

edx_lint/pylint/annotations_check.py
```diff
@@ -20,10 +21,25 @@ def register_checkers(linter):
     Register checkers.
     """
     linter.register_checker(FeatureToggleChecker(linter))
+    linter.register_checker(CodeAnnotationChecker(linter))
```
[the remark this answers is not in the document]
regisb: I think I prefer to insert checkers in the order in which they are defined below.

edx_lint/pylint/annotations_check.py
regisb: > I'm not clear why this couldn't have defined ids ending in 7x, and left all the ids below as-is?

It certainly is possible. There is no perfect approach here, so we just have to make a choice... I suggested what I thought might be the best solution, but I don't have any insight into how numerical IDs are used in CI, so I'm willing to switch to a ordering.

CHANGELOG.rst
regisb: I _think_ that new checks should not be considered as breaking changes... But I'm not sure.

regisb:
I updated the PR. The policy regarding pylint IDs reflects my own understanding of the problem, and may be somewhat limited. If you would like to use a different approach, please say so and I'll implement it.

jmbowman:
I think the warning IDs and names are interchangeable in most contexts, and we definitely prefer to use the names for readability reasons.  It wouldn't surprise me if we use the IDs somewhere (I recently asked for changes on a PR that was about to do so), but I think the odds that we're explicitly referring to these particular ID strings are pretty low.

CHANGELOG.rst
robrap: Technically, I agree with you. From the perspective of someone pulling in an upgrade and wondering whether it might break their build, I think it would be nice to warn them. Given that this already is a major version bump, they will likely read the change log, and they may just concentrate on the `BREAKING CHANGE` section.

Here is a proposed update:
```
* BREAKING CHANGES:
  
  * modify the numerical ID of annotation checks
  * though technically not a breaking change, the new annotation checks may break your build if there are pre-existing violations.
```

edx_lint/pylint/annotations_check.py
robrap: From Jeremy's comment, it sounds like these IDs are not a big deal, so I'm not going to sweat it.  Thanks.  You can mark resolved.

robrap (approved):
Thanks @regisb. One minor proposal for the changlog. Let me know when you want me to merge/release. Thanks!

regisb:
This should be ready to release now @robrap.

PR #143, opened 2021-02-10
BOM-2341 : Don't put amnesty on line continuation

mzulqarnain1:
edx-lint amnesty breaks when it runs into line continuations because it places the comment after the backslash.

Commit for reference:
https://github.com/edx/edx-platform/pull/26391/commits/4afb7c4dfc7005125f4abb3f3440e6b035d27559#diff-9105611e27bef349fe3153c23fdb4789e8c23f1503202e03e19bf14ef70415faR224

Went from:
```
log.warning('Error in conditional module: \
required module {module} has no {module_attr}'.format(module=module, module_attr=attr_name))
```
to
```
 log.warning('Error in conditional module: \  # lint-amnesty, pylint: disable=logging-format-interpolation 
     required module {module} has no {module_attr}'.format(module=module, module_attr=attr_name))
```
which becomes a syntax error.

This PR fixes the problem, now the amnesty will place the comment at the end of line properly like this
```
 log.warning('Error in conditional module: \ 
     required module {module} has no {module_attr}'.format(module=module, module_attr=attr_name)). # lint-amnesty, pylint: disable=logging-format-interpolation
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,12 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[4.1.1] - 2021-03-16
+~~~~~~~~~~~~~~~~~~~~
+
+* Fixed lint amnesty breakage on line continuation
+
+
 [4.1.0] - 2021-02-24
 ~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


UsamaSadiq (approved):

jmbowman (approved):
It looks like this should handle cases where the following line also ends in a backslash, but you may want to keep an eye out for that if you haven't already tested such a scenario.

mzulqarnain1:
> It looks like this should handle cases where the following line also ends in a backslash, but you may want to keep an eye out for that if you haven't already tested such a scenario.

Yes. It works fine. I've already tested that scenario.

PR #144, opened 2021-02-11
Added unittest assertion checker -- BOM-2339

iamsobanjaved:
JIRA:  https://openedx.atlassian.net/browse/BOM-2339

### DIFF
--- a/test/plugins/test_unittest_assert_check.py
+++ b/test/plugins/test_unittest_assert_check.py
@@ -1,0 +1,101 @@
+"""Test unittest_assert_check.py"""
+
+import pytest
+
+from .pylint_test import run_pylint
+
+
+def test_bad_asserts():
+    source = """\
+        import unittest
+
+        class TestUnittestAssertions(unittest.TestCase):
+            def test_wrong_usage(self):
+                # This is the usage of unittest assert functions which shouldn't be used.
+                self.assertEqual('foo'.upper(), 'FOO')
+
+                true = True
+                self.assertTrue(true)
+                self.assertFalse(not true)
+
+                self.assertIn("a", "lala")
+                self.assertNotIn("b", "lala")
+
+                self.assertGreater(1, 0)
+                self.assertLess(1, 2)
+        """
+    messages = run_pylint(source, "avoid-unittest-asserts", '--load-plugins=edx_lint.pylint.unittest_assert')
+    assert messages
+
+
+def test_good_asserts():
+    source = """\
+        import unittest
+
+        class TestPytestAssertions(unittest.TestCase):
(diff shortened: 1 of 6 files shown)


edx_lint/pylint/unittest_assert_check.py
```diff
@@ -0,0 +1,107 @@
...
+        "assertIsNone": "assert arg1 is None",
+        "assertIsNotNone": "assert arg1 is not None",
+        "assertIsInstance": "assert isinstance(arg1, arg2)",
+        "assertNotIsInstance": "assert not isinstance(arg1, arg2)",
+        "assertLess": "assert arg1 < arg2",
+        "assertLessEqual": "assert arg1 <= arg2",
+        "assertGreater": "assert arg1 > arg2",
+        "assertGreaterEqual": "assert arg1 >= arg2",
+        "assertAlmostEqual": "assert round(arg1 - arg2, 7) >= 0",
+        "assertNotAlmostEqual": "assert round(arg1 - arg2, 7) != 0",
```
nedbat, line 72: The math module has [isclose](https://docs.python.org/3/library/math.html#math.isclose) which would be a better replacement.

test/plugins/test_unittest_assert_check.py
```diff
@@ -0,0 +1,101 @@
...
+
+import pytest
+
+from .pylint_test import run_pylint
+
+
+def test_bad_asserts():
+    source = """\
+        import unittest
+
```
mraarif, line 11: maybe move this to a `.py` file and provide a path to that file instead

iamsobanjaved: This is the trend in entire `edx-lint/test/plugins/` directory.

https://github.com/edx/edx-lint/tree/38a1b8580e8621629422cf7fd57d7399c9587bf0/test/plugins

mraarif: it shouldn't be hard to read from a file instead, it'd be a lot cleaner solution

edx_lint/pylint/unittest_assert_check.py
iamsobanjaved: Replaced with `isclose`

test/plugins/test_unittest_assert_check.py
nedbat: An advantage of keeping the code in a string is that I can read and understand the whole test without having to go find another file.

iamsobanjaved: I have moved the test code back to strings

jmbowman (approved):
One wording improvement suggestion, otherwise looks good to me.

PR #146, opened 2021-02-23
Remove base django requirement

UsamaSadiq:
**JIRA Issue:** [BOM-2418](https://openedx.atlassian.net/browse/BOM-2418)
#### Details: 
Removing the `Django` requirement from `base.in` because: 
- `Django` requirement in `base.in` causes the `django` version in [testeng-ci repo](https://github.com/edx/testeng-ci/blob/981773e6a419cc8347183624893c11cde11f93c0/requirements/testing.txt#L50) to upgrade to `django==3.1.6` which causes failures in different [jobs](https://build.testeng.edx.org/job/travis-job-counts/259251/console). 
- It upgrades the Django version to 3+ in any repo that uses this package without applying constraint on `Django` versions.

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -3,7 +3,6 @@
 code-annotations
 click
 click-log
-Django
 pylint
 pylint-django
 pylint-celery
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -4,7 +4,7 @@
 #
 #    make upgrade
 #
-astroid==2.4.2
+astroid==2.5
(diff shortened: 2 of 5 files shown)


jmbowman (approved):
This is a good start, but doesn't fully resolve the original ticket due to the remaining code-annotations dependency on Django.  See https://github.com/edx/code-annotations/pull/64#issuecomment-799668163 for suggested next steps on fixing that.

PR #147, opened 2021-03-11
adr and how-to for new lint checks

robrap:
See newly added docs for details of the best practice for
introducing new lint checks.


docs/decisions/0001-adding-new-checks.rst
```diff
@@ -0,0 +1,36 @@
...
+Decisions
+---------
+
+* New checks should be introduced as optional, even if you ultimately want them to be required by default.
+
+  * For rolling out new required-by-default checks, making checks optional at first enables repo owners to fix or amnesty violations on their own timeline before they become required, or batch fixing or amnestying larger batches of checks when they become required.
+  * See `how_tos/0001-adding-new-checks.rst` for details in making this happen.
+
+* Mark new checks that will become required as deprecated as disabled/optional checks.
+* Communicate new disabled/optional checks that will be required-by-default in the future.
```
robrap, line 28: **Reviewer Note:** I did not yet clarify where.  What should we note here?

docs/how_tos/communicate-new-checks.rst
```diff
@@ -0,0 +1,53 @@
...
+* Use a major version bump to communicate breaking changes.
+* Follow the instructions in the pylintrc file for making a "CENTRAL CHANGE".
+
+  * Move optional checks from the disable list to the enable list in pylintrc.
+
+* Add a CHANGELOG.rst entry noting the breaking change(s).
+
+  * List the checks that are now required-by-default.
+  * Note that violations of the new checks can either be fixed, or amnestied following instructions in the README.
+  * If instructions for fixing the new checks is not clear from the check's message, add a link to any additional documentation required to fix.
```
robrap, line 53: Note to self: Duplicate or more the note on how to fix the violation up earlier to the changelog entry for first introducing the check as disabled/deprecated.

docs/how_tos/communicate-new-checks.rst
```diff
@@ -0,0 +1,53 @@
...
+
+* Follow the instructions in the pylintrc file for making a "CENTRAL CHANGE".
+
+  * Add the plugin to ``load-plugins`` as detailed above in pylintrc.
+  * Ensure all new checks are added under the disable list, and not the enable list in pylintrc.
+
+* Add the new CHANGELOG.rst entry:
+
+  * Explain the purpose of the new plugin and list its new checks.
+  * List any new checks that will become required-by-default under a "Deprecated" section, noting the fact that they are disabled is deprecated, and they will be required/enabled by default, in a future major release.
```
robrap, line 28: **Note to self:** Probably should also explain how to make "LOCAL CHANGE" in pylintrc to enable locally in advance of it becoming required.

docs/decisions/0001-adding-new-checks.rst
```diff
@@ -0,0 +1,36 @@
...
+
+* What changes should be considered a breaking change, requiring a major version bump.
+* How to avoid too avoid many breaking changes.
+* When and where to best communicate changes.
+* How to ensure that new violations are being picked up by repositories.
+
+Decisions
+---------
+
+* New checks should be introduced as optional, even if you ultimately want them to be required by default.
```
nedbat, line 22: Am I forgetting something? What's an optional pylint check? Or is this for checks beyond pylint?

docs/decisions/0001-adding-new-checks.rst
```diff
@@ -0,0 +1,36 @@
...
+
+  * For rolling out new required-by-default checks, making checks optional at first enables repo owners to fix or amnesty violations on their own timeline before they become required, or batch fixing or amnestying larger batches of checks when they become required.
+  * See `how_tos/0001-adding-new-checks.rst` for details in making this happen.
+
+* Mark new checks that will become required as deprecated as disabled/optional checks.
```
nedbat, line 27: I don't understand this sentence.

docs/decisions/0001-adding-new-checks.rst
```diff
@@ -0,0 +1,36 @@
...
+* Communicate new disabled/optional checks that will be required-by-default in the future.
+* Use a major version bump when making a group of checks required-by-default. The major version should be enough for communication.
+* Add automation to upgrade the pylintrc file in each repo with changes made in edx-lint.
```
nedbat, line 30: A link to example automation would be great.

docs/how_tos/communicate-new-checks.rst
```diff
@@ -0,0 +1,53 @@
...
+==============================================
+
+Adding a new lint plugin
+------------------------
+
+After coding a new lint plugin, it should start off being optional, even if you intend to make it required-by-default in the future. See decisions/0001-adding-new-checks.rst for background on why.
+
+If all checks in the plugin will always be optional, simply provide instructions in the CHANGELOG.rst for how to add the plugin. For example::
+
+  * Explain the purpose of the new plugin, and potentially list its new checks.
```
nedbat, line 11: This is an instruction rather than an example.

docs/how_tos/communicate-new-checks.rst
nedbat, line 28: I find this confusing: "the fact that they are disabled is deprecated."  I wouldn't describe a change in a setting as deprecating the old value of the setting.  Could we instead say, "Breaking change: a previously disabled opt-in check is now required by default."

docs/decisions/0001-adding-new-checks.rst
robrap: I wasn't sure what I meant when I first wrote this.  What I now think is that checks will be disabled by default (i.e. optional, or opt-in using a local override to enable), and eventually will be enabled by default (i.e. required by default, or opt-out using a local override to disable).

First, can you confirm that this is a reasonable plan? Also, is it more clear to lead with pylint language like "enabled/disabled" vs "required/optional"? Should I drop the language of "required/optional" altogether?

docs/decisions/0001-adding-new-checks.rst
robrap: Me neither. 😄 I was rushing on the ADR, trying to balance not repeating too much of the how-to, but also providing a summary of decisions that might change. Once the rest is cleaned-up, I will fix this.  Thanks.

docs/decisions/0001-adding-new-checks.rst
robrap: That will have to come later when it exists.  Creating this is part of the "Consequences".

docs/how_tos/communicate-new-checks.rst
robrap: Thanks.  I'll fix.  I reformatted and combined sections, and  looks like I didn't fully fix this.

docs/how_tos/communicate-new-checks.rst
robrap: Yes.  Also, see other comment on language.  Might even end your sentence with "...enabled by default".

docs/decisions/0001-adding-new-checks.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Draft
+
+Context
+-------
+
+When adding new violations to edx-lint, there currently isn't clarity around the following:
+
+* What changes should be considered a breaking change, requiring a major version bump.
+* How to avoid too avoid many breaking changes.
```
regisb, line 15: "How to avoid too ~~avoid~~ many breaking changes."

docs/decisions/0001-adding-new-checks.rst
regisb: I'm not sure this is possible without modifying the pylintrc file for every project. This is because all linters from edx_lint.pylint are run by default.

regisb (commented):
My problem here is that this ADR seems to rest on the assumption that it's somehow possible to make optional lint checkers, but I have no idea how to do that. Whenever we run pylint, we include all linters from edx_lint.pylint. Excluding some would require to list them explicitely in every project's pylintrc.

docs/decisions/0001-adding-new-checks.rst
robrap: No longer applicable.

docs/how_tos/communicate-new-checks.rst
robrap: Added instructions to the README.

docs/how_tos/communicate-new-checks.rst
robrap: Not applicable, and updated README.

docs/decisions/0001-adding-new-checks.rst
robrap: @regisb @nedbat: Thanks for your feedback. A combo of your notes, and doing some research, helped me understand what is going on.
1. I simplified the docs, because in most cases, we are stuck with breaking changes.
2. I added some more vague notes about optional plugins, because it is an option, but I don't have enough experience with this to recommend it, etc.

docs/decisions/0001-adding-new-checks.rst
robrap: @nedbat: Let me know if this good enough, or if you have a suggestion for clarifying in the text?  Should I add a note that we should add a link once it has been implemented?  Maybe this should not be in the ADR at all at this point?

docs/how_tos/communicate-new-checks.rst
robrap: This section has been removed/rewritten.

docs/how_tos/communicate-new-checks.rst
robrap: All of this has changed.

robrap:
@regisb @nedbat: I did a pretty major overhaul based on feedback and better understanding what is possible.  See https://github.com/edx/edx-lint/pull/147#discussion_r594591599 for more info.

jinder1s (approved):

PR #151, opened 2021-04-02
feat: added re-useable github action for commmit linting

iamsobanjaved:


feanil:
This is now handled via the [repo-checks](https://github.com/openedx/terraform-github/blob/main/migrate/repo_checks.py) mechanism and so we don't need it to be part of edx-lint.

PR #177, opened 2021-07-29
feat: Adding a constraint for social-auth-core that is caused by other common constraints

staubina:
The constraints that block us from upgrading social-auth-core are part of the common constraints file. I am moving the social-auth-core constraint here as well to make it clear that there is a relationship.


UsamaSadiq (approved):
Looks Okay to merge. `PyJWT` constraint will soon be removed for `Django32` upgrade so this constraint will also be updated.

crice100 (approved):

PR #196, opened 2021-09-17
feat: a new "update" command.

nedbat:
The "update" command will write any file that already exists.  It still checks that the existing file is a valid edx_lint-written file, so it won't destroy local edits.

### DIFF
--- a/test/test_cmd.py
+++ b/test/test_cmd.py
@@ -4,6 +4,7 @@ import os
 import shutil
 import tempfile
 import unittest
+from unittest.mock import patch
 
 from edx_lint.cmd import main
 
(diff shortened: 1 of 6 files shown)


nedbat:
This is ready for review.

test/test_cmd.py
```diff
@@ -109,3 +116,42 @@ def test_file_types(self):
...
+    def test_update_wont_clobber(self):
+        # Create a hand-written pylintrc file:
+        with open(PYLINTRC, "w") as pylintrc:
+            pylintrc.write("[bogus]\nnot_really = pylintrc\n")
+
+        # Update will see it's not one made by edx-lint, and move it aside.
+        self.assertEqual(0, self.call_command(["update"]))
+
+        # The backup file is our original hand-written file.
+        self.assert_file(PYLINTRC_BACKUP, contains="not_really = pylintrc")
```
feanil, line 155: Should this be a check in the original file if we want to test that it wasn't clobbered?

feanil (commented):
Generally looks good but I don't quite understand one test.

test/test_cmd.py
nedbat: edx-lint checks the existing file.  If it wasn't written by edx-lint, then it moves the existing file to a backup, and then writes the new file.  So when write (or update) is done, the real file is always the recent edx-lint output.  "Not clobbering" the hand-written file might be a misnomer.  The data hasn't been lost, but it has been shoved aside.

feanil: Got it so it's more like `test_update_will_backup`.  Please add your reply or something like it as a comment in this test function.

feanil (approved):
One request to add an additional comment but then it looks good to go after that.

PR #199, opened 2021-09-20
feat: edx-lint can write commitlint.config.js files

nedbat:
This is to support the upcoming GitHub Action to check conformance to OEP-51.


edx_lint/files/commitlint.config.js
```diff
@@ -0,0 +1,27 @@
...
+    'footer-leading-blank': [0, "always"],
+    'subject-case': [0, "always", []],
+    'subject-full-stop': [0, "never", '.'],
+  },
+
+  ignores: [
+    // Allow GitHub revert messages, like:
+    //    Revert "introduce a bug"
+    //    Revert "introduce a bug" (#1234)
+    message => /^Revert ".*"( \(#\d+\))?/.test(message),
```
feanil, line 25: Non-blocking suggestion: I mentioned this on the webhooks PR, but it may be worth adding an ignore for the `Merge pull request...` message as well to handle the case where people merged PRs into other PRs before landing the whole change upstream.

feanil (approved):

edx_lint/files/commitlint.config.js
feanil: Ah nevermind, I see your response over there.  commitlint has a default that already ignores that one but not revert messages.

nedbat:
@feanil Could be worth a last look.

feanil (approved):

PR #215, opened 2021-11-03
fix: warn on "squash" commit type

robrap:
**Description:**

Warn on "squash" commit type

Note: This might not work for "squash! ..." git autosquash commit messages, because it can't be marked as an acceptable type.

**Testing**

Note: Couldn't get local testing working, so I just used this PR and the github action to test.

While having "squash" commits on the PR, I got a passing build with the following warnings:
> You have commit messages with warnings
> 
> ⧗   input: squash: remove failOnWarnings
> ⚠   Reminder to squash all 'squash' commits [squash-match]
> 
> ⚠   found 0 problems, 1 warnings

**Merge checklist:**
- [ ] All reviewers approved
- [x] CI build is green
- [x] If adding new checks, followed how_tos/0001-adding-new-checks.rst
- [x] Changelog record added (if needed)
- [x] Documentation updated (not only docstrings) (if needed)
- [ ] Commits are squashed


edx_lint/files/commitlint.config.js
```diff
@@ -6,9 +6,13 @@ module.exports = {
   rules: {
     'type-enum':
       [2, 'always', [
-        'revert', 'feat', 'fix', 'perf', 'docs', 'test', 'build', 'refactor', 'style', 'chore', 'temp',
+        'revert', 'feat', 'fix', 'perf', 'docs', 'test', 'build', 'refactor', 'style', 'chore', 'temp', 'squash',
       ]],
-
+    'type-enum':
```
inventhouse, line 11: I don't _think_ duplicate keys are allowed.

robrap: I also am skeptical. But this also shows the approach we'd probably need to take with your plugin idea. In other words, "squash" would need to be added to the accept list, and we'd need a custom rule that `never` allows a message that starts with "squash", but is just a warning.

inventhouse: Yes, exactly, but since it's a dictionary it needs to have a different name; hence the need for a custom rule.

robrap: Ah yes. I was mistakenly thinking it was up to their implementation internally whether or not they accepted the same key twice, but ignoring the fact that these rules are using an object and not a list, and this clearly won't work as you stated. And yes - a custom rule provides a new key - which is what we need.

edx_lint/files/commitlint.config.js
```diff
@@ -28,4 +29,22 @@ module.exports = {
...
+
+  plugins: [
+    {
+      rules: {
+        'squash-match': ({subject}) => {
+          const SQUASH = 'squash';
+          return [
+            // matches squash commits, so use with "never"
+            // - Why does this match on empty type/subject?
+            // - Gives a false warning in this case, but not a big deal.
```
robrap, line 41: **Note to Reviewer:** Using a commit with a subject, but no type at the beginning (e.g. "what is wrong with this commit"), the squash warning fires, but so does an empty subject error (which is not correct), in addition to the empty type error (which is correct).

edx_lint/files/commitlint.config.js
robrap: PR has been updated.

robrap:
Since `commitlint` already ignores `squash!` and `fixup!`, I'm going to close this PR rather than adding custom rules.

PR #238, opened 2022-01-18
fix: add common constraint of redis<4

iloveagent57:


gabor-boros:
@robrap @iamsobanjaved

This change blocks https://github.com/openedx/edx-platform/pull/29046 on my side as the linked PR created to be able to upgrade celery **and** use Redis ACLs (introduced in Redis 6). The ACL support was introduced in `redis-py` version 4.x. 

I checked Redis server documentation about releases ([ref](https://redis.io/topics/releases)) which says:

> As a rule, older versions are not supported as we try very hard to make the Redis API mostly backward compatible.
> Upgrading to newer versions is the recommended approach and is usually trivial.
> The latest stable release is always fully supported and maintained.
> Two additional versions receive maintenance only, meaning that only fixes for critical bugs and major security issues are committed and released as patches:
>     - The previous minor version of the latest stable release.
>     - The previous stable major release.
> For example, consider the following hypothetical versions: 1.2, 2.0, 2.2, 3.0, 3.2, ...
> When version 2.2 is the latest stable release, both 2.0 and 1.2 are maintained.
> Once version 3.0.0 replaces 2.2 as the latest stable, versions 2.0 and 2.2 are maintained, whereas version 1.x reaches its end of life.
> This process repeats with version 3.2.0, after which only versions 2.2 and 3.0 are maintained.
> The above are guidelines rather than rules set in stone and will not replace common sense.

Based on that, the supported versions are `6.2`, `6.0`, `5.0`. 

@iloveagent57 Why was this change necessary? Is Redis server 3 still in use at edX?
@robrap @iamsobanjaved How could we resolve this issue? Should I create a revert PR on this change? I have the feeling that supporting unsupported Redis versions in edx-platform is not what is expected.

robrap:
@gabor-boros: We've had some recent celery related deployment issues. This constraint was put in place in case it was related to this upgrade, although that hypothesis is in doubt. There is active and ongoing work to get to the bottom of this, and @iloveagent57 can provide more information on timeline for that. Also, it may be that we could remove the constraint in advance of getting to the bottom of the problem, but @iloveagent57 will again need to speak to that.

iloveagent57:
@gabor-boros and @robrap I believe it's safe to remove this pin, sorry for the confusion.  I've opened a revert PR that contains a little more context: https://github.com/openedx/edx-lint/pull/241

PR #279, opened 2022-08-26
fix: add desc for setuptools pin

aht007:
Add desc for `setuptools` pin


awais786 (approved):

nedbat:
Thanks.  That setuptools bug report (https://github.com/pypa/setuptools/issues/2940) says it was fixed within a day.  Do we think we can remove this pin now?  Do we know which services were affected so we can test them?

aht007:
It wasn't the only bug in `setuptools==60.0` as Jeremy mentioned here. But last I checked version 65 had been released on PyPI so essentially all the issues should have been resolved by now. We don't know about which services were affected as we don't have the slack messages from that time around but I will create a ticket in our board to investigate and remove this constraint.
Screen Shot 2021-12-23 at 12 43 32 PM

PR #282, opened 2022-09-07
Fix github url strings (org edx -> openedx)

sarina:
## This PR was autogenerated

This pr replaces the old GitHub organization, github.com/edx, with the new GitHub organization, github.com/openedx.

Tagging @openedx/tcril-engineering for review, but others are welcome to provide review.

Ref: https://github.com/openedx/tcril-engineering/issues/42


nedbat:
You need to run edx-lint to write the files in the root of the directory. There are instructions in the file about what to do.  Instead of editing those files, update write.py (as you have done), then run `edx-lint write .editorconfig` (etc) to update the .editorconfig file.

nedbat:
Now this PR doesn’t update the .editorconfig file itself in edx-lint. Is that intentional? I can understand if so: get a new edx-lint, publish it, and then use it on everything, including itself?

sarina:
@nedbat - the build is now failing due to [two instances of R201](https://github.com/openedx/edx-lint/runs/8235130294?check_suite_focus=true), R0201: Method could be a function (no-self-use).

Can I ignore these? Is this a helpful warning for the project?

nedbat:
Yes, we can silence those violations, but it's not easy.  See #283.

sarina:
That worked, thanks @nedbat 

Could I get a re-review from you or @feanil ?

edx_lint/pylint/module_trace.py
```diff
@@ -41,6 +41,7 @@ class ModuleTracingChecker(BaseChecker):
 
     def visit_module(self, node):
         """Called for each module being examined."""
+        self    # pylint: disable=pointless-statement
```
feanil, line 44: Is there a reason this pointless statement is being left here? Can that reason be left in a comment next to this?

test/test_cmd.py
```diff
@@ -42,6 +42,7 @@ def call_command(self, argv=None):
         Arguments:
             argv (list) -- arguments to pass to the edx_lint script
         """
+        self    # pylint: disable=pointless-statement
```
feanil, line 45: Same as the comment above, if this is not useless, A comment would help and if it is useless, then removing the line would be good.

feanil (changes requested):

edx_lint/pylint/module_trace.py
sarina: @nedbat did this to silence the R201 warning, see https://github.com/openedx/edx-lint/pull/283

feanil (commented):
Weird that we didn't remove the lines but let's unblock this for now and we can deal with that as a follow up.

feanil (approved):

sarina:
Thanks! I'll create a tag for `5.2.5` now, hoping that triggers a release.

PR #285, opened 2022-09-12
style: Remove pointless statements.

feanil:


feanil:
Turns out this is a rabbit hole.  This issue is fixed in a newer version of pylint where this particular check is dropped. But currently we're trying to make pylint pass on both an old and new version which results in a situation where fixing one breaks the other.  The `self` call + disable is the only way to make them both work.  I put the "pointless statements" back and added a comment.

nedbat (approved):

PR #298, opened 2022-11-30
fix: pylint<2.15.0 check removed

zubairshakoorarbisoft:
Just removed pylint constraint and added findings as this PR is not ready to merge yet.
Issue: https://github.com/orgs/edx/projects/12/views/1?filterQuery=293

So, finally i just come up with the findings that, the new pylint version `v2.15.7` is not supporting the validations exactly for which we have the tests in place.
Screenshot 2022-12-05 at 11 04 34 PM

Like we have test here to check there should be an error where we are not calling super method from its child/drived class.
but if you can see in the output of all possible pylint error messages, no one contains the info like this which is checking super method is calling or not.


requirements/constraints.txt
```diff
@@ -5,4 +5,4 @@ code-annotations>=1.1.0
 
 # pylint>=2.15.0 contains breaking changes
```
UsamaSadiq, line 6: We should remove this comment completely from the constraints file.

tox.ini
```diff
@@ -5,7 +5,7 @@ envlist = py38, coverage, pylint, pylint212
 deps =
     -r{toxinidir}/requirements/test.txt
 commands =
-    coverage run -p -m pytest {posargs:}
+    coverage run -p -m pytest {posargs:} -vv
```
UsamaSadiq, line 8: If this flag was added only for debugging, then it should be removed.

UsamaSadiq (commented):
Once this PR is merged, we should also create a follow up PR to bump and release next version of `edx-lint` so all the other packages can also start running pylint checks with the new version.

PR #319, opened 2023-02-15
fix: Be precise about the exception we're ignoring.

feanil:
We're currently getting the following in pylint output:

pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.

### DIFF
--- a/edx_lint/files/pylintrc
+++ b/edx_lint/files/pylintrc
@@ -509,4 +509,4 @@ int-import-graph=
 
 # Exceptions that will emit a warning when being caught. Defaults to
 # "Exception"
-overgeneral-exceptions=Exception
+overgeneral-exceptions=builtins.Exception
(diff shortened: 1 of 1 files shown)


nedbat (approved):

UsamaSadiq (approved):

edx_lint/files/pylintrc
```diff
@@ -509,4 +509,4 @@ int-import-graph=
 
 # Exceptions that will emit a warning when being caught. Defaults to
 # "Exception"
```
UsamaSadiq, line 511:
```suggestion
# "builtins.Exception"
```

PR #349, opened 2023-06-20
build: Remove constraint on sphinx<6.0.0

feanil:
We no longer need this constraint because non of the repositories in
openedx use the `edx-sphinx-theme`.  The theme and extensions we use now
properly advertise their compatability and are maintained by a 3rd
party (sphinx-book-theme).


timmc-edx (approved):
Looks like there are 20 repos left in the edx org to update. I'm guessing we have an ongoing project at 2U to clean these up. (None appear to be critical-path, high-update-frequency repos.)

Issue #336, opened 2023-05-03
Unpin tox


jmbowman:
We [pinned tox](https://github.com/openedx/edx-lint/blob/master/edx_lint/files/common_constraints.txt#L25-L27) back in December because the 4.0.0 release broke most of our CI runs.  It's quite possible that this has been resolved by now, so we should attempt to remove this pin (requested by Axim in https://openedx.slack.com/archives/C0497NQCLBT/p1682004301805179).

Proposed A/C:
* Use the workaround from the Slack thread above to allow upgrading tox in edx-platform (chosen because it's the repo most likely to hit nasty corner cases in almost any upgrade).  If CI fails with the latest version, spend a little time attempting to resolve.
* If we can get edx-platform CI working with the latest tox release, merge the PR with those changes.  If any changes were needed which are also likely to be needed in other repos upgrading to the latest version, announce them via all appropriate communications channels.
* If all the above goes well, remove the pin from `common_constraints.txt` so other repos can upgrade tox normally.
* Remove the edx-platform workaround for the common tox constraint, if it was merged.

jmbowman:
Another reason to get this done: the old version of tox depends on the `py` package, which has an outstanding security flaw.  It doesn't impact us (only exploitable if used in conjunction with a Subversion repository), but it clutters up the security dashboard with issues we'll otherwise need to manually inspect and dismiss in dozens of repositories.

iamsobanjaved:
Need the following issue to be done, before we can remove the tox constraint from common constraints. Because `whitelist_externals` is removed from tox version `4.0.0`. While we can do the first step mentioned in the description, unpin explicitly in edx-platform by removing tox constraint using `sed` command in make upgrade.

- https://github.com/edx/edx-arch-experiments/issues/130

PR #354, opened 2023-07-12
build: Remove tox constraint.

feanil:
The `tox` package is already 6 minor revisions ahead at 4.6.4.  If there
are still plugins that don't support 4.x.x, then they are likely
stagnant and need to be removed or updated ourselves.

However, as long as we keep this constraint here, we can't easily find
and fix those issues.  In many of the cases, this constraint was added
due to the incompatibility of tox-battery with tox 4.x.x.  However,
tox-battery has updated its `install_requires` to be explicit of this
dependency.

https://github.com/signalpillar/tox-battery/blob/master/setup.py#L20

Another issue we're running into is that some of the dependencies of tox
are starting to publish security vulnerabilities.  It's lower risk since
this is in dev and CI but leaving this as is will increase security
noise making it harder to respnod to real signals.

Specifically, tox<4.0.0 depends on a version of `py` which has a
security vulnerability.  Dependabot is picking this up and making some
noise in a lot of our repos.

Closes https://github.com/openedx/public-engineering/issues/203

### Process Note
Before merging this, I'll be posting a detailed message on Discourse(cross posted to slack) to let people know it's coming, what to look for and how to fix it locally in their repos.  That should allow people to fix forward locally without the need for the global constraint.  This will also allow us to incrementally move forward and fix one or two repos at a time.


edx_lint/files/common_constraints.txt
```diff
@@ -21,7 +21,3 @@ elasticsearch<7.14.0
 
 # django-simple-history>3.0.0 adds indexing and causes a lot of migrations to be affected
 django-simple-history==3.0.0
-
-# tox>4.0.0 isn't yet compatible with many tox plugins, causing CI failures in almost all repos.
```
robrap, line 25: > ... causing CI failures in almost all repos.

This scares me, and seem like a good reason to have a common constraint. Did you confirm that a repo that was failing on 4.0.0 now works with the latest upgrade, and that we can reasonably expect that it is now passing for most repos?

feanil: Unfortunately, it's nearly impossible to test this because of how the common constraints are setup in a repo.  You can't actually remove the constraint at the repo level without `make upgrade` automatically adding it back before it rebuilds requirements.

That said, this could be added to the constraints at the repo level and my plan was to do a big announce so that maintainers are prepared to do that if they see issues.  As I said in the description, most of our repos have `tox-battery` as a dependency and that already constrains `tox<4.0.0` so I'm less concerned about this breaking everything like it did when tox 4.0.0 was first released. (At that time tox-battery did not have its constraint on `tox<4.0.0`.

robrap: We could use this hacky solution on a branch just for testing: https://github.com/openedx/edx-platform/commit/56d12bf230bc1fc5ec930c3afdf412ba2cae7239. I think it would be useful. If we discover a problem that is easily fixable, but that all or many repos might run into, it might be nice to provide the solution rather than having everyone add constraints, and then having the same problem we have now (a constraint where we don't want one), only a much more global version of it. Thoughts?

robrap: Also see https://github.com/openedx/edx-lint/issues/355 which I just moved from Jira: https://github.com/openedx/edx-lint/issues/355.

feanil: @robrap good call on digging further into this. I had missed the fact that 0.6.2 of tox-battery(the version that constrains tox) has been merged but has not been released.  I think the real path forward is gonna have to be dropping the dependency on tox-battery everywhere first and then dropping this common constraint.

robrap: Cool. Thanks for digging further. Good luck. Should this ticket be closed for now, or will you block on some new issue?

robrap: Sounds good. Are you planning on ticketing and pushing forward the `tox-battery` work? Was this all just inform for me at this point? Thanks.

feanil:
FYI, there is now a new release of tox-battery with the constraint properly declared so this should be safe to remove.  I've

iamsobanjaved:
We have[ an issue on our board for this](https://github.com/openedx/edx-lint/issues/336) and was pending due to the team focusing on the Django upgrade, but now we have prioritized this and will work on first upgrading tox in edx-platform and then remove this constraint. Also this constraint removal work is blocked this issue https://github.com/edx/edx-arch-experiments/issues/130

feanil:
This got fixed in a different PR.

PR #398, opened 2024-02-15
chore: Updates to support Python up to 3.12

### LINKED ISSUE #397: Errors in Python 3.12
While attempting to upgrade a project to Py3.12 I started getting this error in quality checks:

`Command line or configuration file:1:0: E0013: Plugin 'edx_lint.pylint' is impossible to load, is it installed ? ('No module named 'pkg_resources'') (bad-plugin-value)`

Older versions of Python passed as usual. It looks like pkg_resources has been superseded by [importlib.resources](https://docs.python.org/3/library/importlib.resources.html), at least for the calls we're using it for in this repository. It should support back to Python 3.7. 

I'm hoping to have a PR for this in the near term, but in case someone runs into this sooner it should be an easy fix. I don't know if there are other 3.12 issues behind it, though.

bmtcril:
**Description:**

CI and fixes to support Python 3.8 - 3.12.

Closes #397 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] If adding new checks, followed how_tos/0001-adding-new-checks.rst
- [ ] Changelog record added (if needed)
- [ ] Documentation updated (not only docstrings) (if needed)
- [ ] Commits are squashed


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38, coverage, pylint
+envlist = py{38,39,310,311,312}, coverage, pylint
```
UsamaSadiq, line 2: We are only using py38 and py312 in CI so we can skip rest of the versions.
```suggestion
envlist = py{38,312}, coverage, pylint
```

UsamaSadiq (changes requested):
We don't want to upgrade the requirements for now since it may break services with updated package dependencies. 
We plan to update requirements upgrade workflows to run on Python 3.12 after we've completed edx-platform upgrade to 3.12 to avoid any major issue.

PR #435, opened 2024-07-15
feanil/remove custom lint config

feanil:
- **build: Remove the custom commitlint file.**
- **feat!: Don't allow writing of commitlint.config.js files.**


bmtcril (approved):
🙌

Issue #476, opened 2025-01-07
Unpin urllib3


mumarkhan999:
**Related issue:**

- https://github.com/openedx/edx-lint/issues/475

Issue #475, opened 2025-01-07
Pin urllib3 in global constraints


mumarkhan999:
**Description**

The new version of urllib3 isn't compatible with httppretty and causing the following errors

```
E           httpretty.errors.UnmockedError: Failed to socket.shutdown because because a real socket does not exist
E           
E           Tip: You could try setting (allow_net_connect=True) to allow unregistered requests through a real TCP connection in addition to (verbose=True) to debug the issue.
```

More details: https://github.com/gabrielfalcao/HTTPretty/issues/484

PR #500, opened 2025-07-29
Remove some old common constraints

feanil:
- fix: Don't pin django-simple-history

    We removed this pin in edx-platform a while ago and if we can run these
    migrations there, I think any other services should be able to handle
    this as well.  Also there is no ticket with details here and so it is
    not clear if there was a specific concern other than performance issues
    with adding indexes.

    Reference (default branch in edx-platform as of today):
    https://github.com/openedx/edx-platform/blob/d29171c04649f29a26bfc96aa73b64bf01bfdd5b/Makefile#L124-L125

- fix: Drop the urllib3 constraint

    Other updates seem to have resolved the issue reported when this was
    pinned.  The issue there seems to be resolved in edx-platform now as
    shown in https://github.com/openedx/edx-platform/pull/37059

    Given that this was the only known issue with this, we can unpin this
    library.

Resolves https://github.com/openedx/edx-lint/issues/475
Resolves https://github.com/openedx/edx-lint/issues/476

### DIFF
--- a/edx_lint/files/common_constraints.txt
+++ b/edx_lint/files/common_constraints.txt
@@ -25,7 +25,3 @@ django-simple-history==3.0.0
 # Cause: https://github.com/openedx/edx-lint/issues/458
 # This can be unpinned once https://github.com/openedx/edx-lint/issues/459 has been resolved.
 pip<24.3
-
-# Cause: https://github.com/openedx/edx-lint/issues/475
-# This can be unpinned once https://github.com/openedx/edx-lint/issues/476 has been resolved.
-urllib3<2.3.0
--- a/edx_lint/files/common_constraints.txt
+++ b/edx_lint/files/common_constraints.txt
@@ -19,9 +19,6 @@ Django<5.0
 # See https://github.com/openedx/edx-platform/issues/35126 for more info
 elasticsearch<7.14.0
 
-# django-simple-history>3.0.0 adds indexing and causes a lot of migrations to be affected
-django-simple-history==3.0.0
-
 # Cause: https://github.com/openedx/edx-lint/issues/458
 # This can be unpinned once https://github.com/openedx/edx-lint/issues/459 has been resolved.
 pip<24.3


edx_lint/files/common_constraints.txt
```diff
@@ -19,13 +19,6 @@ Django<5.0
 # See https://github.com/openedx/edx-platform/issues/35126 for more info
 elasticsearch<7.14.0
 
-# django-simple-history>3.0.0 adds indexing and causes a lot of migrations to be affected
```
robrap, line 22: 1. This would warrant a note in #cc-risky-changes. That note could go in now with any guess you have about the timeline.
2. @UsamaSadiq made this change, so maybe he can provide more details?
3. We added the following help text in edx-platform constraints, that might be useful in this file as well:
https://github.com/openedx/edx-platform/blob/master/requirements/constraints.txt#L10C1-L11C112
4. I mentioned the proposal for overriding common constraints in the other PR: https://github.com/openedx/edx-platform/pull/37059#discussion_r2240128431.
  - I'm not sure if anyone has time to implement, but it could help with removing common constraints by allowing us to easily test a more up-to-date dependency in a single repo before opening the floodgates for all repos.
  - I understand that this probably isn't going to happen for this particular constraint.

robrap: Great. I didn’t realize the change was already made in the riskiest repo. Sounds like it shouldn’t be an issue then, assuming no manual intervention was required.

feanil:
Thanks for taking a look y'all and I appreciate the input.   I've mentioned this in #cc-risky-changes and as far as I can tell there are no other actionable changes left on this PR (let me know if I missed something).  Can one or both of you approve?

UsamaSadiq (approved):
Approved. Gotta keep an eye on the next edx-platform requirements upgrade PR for assurance.

feanil:
I'll be making one of those manually after this lands and I'll keep an eye on it.

PR #537, opened 2026-04-23
feat: add write_uv_constraints command to manage uv constraint-dependencies

feanil:
## Summary

Adds a new `edx_lint write_uv_constraints` command to support repos migrating from `pip-compile` to `uv`.

**The problem:** `uv lock` has no `--constraint` CLI flag — version constraints must live in `[tool.uv].constraint-dependencies` inside `pyproject.toml`. Without tooling, every repo would need to copy-paste edx-lint's global constraints and keep them in sync manually.

**The solution:** A command that:
- Reads the bundled global constraints from `edx_lint/files/common_constraints.txt`
- Optionally merges in repo-specific constraints from a local file (typically `constraints.txt` next to `pyproject.toml`)
- Writes the merged list into `[tool.uv].constraint-dependencies`, preserving all other `pyproject.toml` formatting and content via `tomlkit`
- Adds a prominent "DO NOT EDIT" comment above the key on first write
- Is idempotent: running it twice produces the same result

**Usage:**
```bash
# Global constraints only
edx_lint write_uv_constraints

# Merge with repo-specific constraints
edx_lint write_uv_constraints pyproject.toml --local constraints.txt
```

The `constraint-dependencies` list is fully machine-managed. Humans should never edit it directly — global changes go in `edx_lint/files/common_constraints.txt`, repo-specific pins go in a local `constraints.txt`.

## Changes

- `edx_lint/cmd/write_uv_constraints.py` — new command module
- `edx_lint/cmd/main.py` — register command in dispatcher and help output
- `requirements/base.in` — add `tomlkit` as a direct dependency (was already present as a transitive dep of `pylint`)
- `test/test_cmd.py` — 10 new tests covering all key behaviours
- `requirements/*.txt` — regenerated via `make upgrade`

### DIFF
--- a/edx_lint/cmd/write_uv_constraints.py
+++ b/edx_lint/cmd/write_uv_constraints.py
@@ -1,0 +1,107 @@
+"""The edx_lint write_uv_constraints command."""
+
+import argparse
+import importlib.resources
+import re
+
+import tomlkit
+
+
+def _package_name(spec):
+    """Return the normalized package name from a constraint specifier like 'Django<6.0'."""
+    name = re.split(r"[<>=!~\[;@]", spec, maxsplit=1)[0].strip()
+    return name.lower().replace("-", "_").replace(".", "_")
+
+
+def _parse_constraints(text):
+    """Parse constraints file text, stripping comments and blank lines.
+
+    Returns a tuple of (constraints, directives) where:
+      constraints: valid package-specifier lines
+      directives:  pip directive lines (-c, -r, etc.) that are not valid here
+    """
+    constraints = []
+    directives = []
+    for line in text.splitlines():
+        line = line.split("#")[0].strip()
+        if not line:
+            continue
+        if line.startswith("-"):
+            directives.append(line)
+        else:
+            constraints.append(line)
+    return constraints, directives
+
+
+def write_uv_constraints_main(argv):
+    """
+    write_uv_constraints [pyproject.toml]
+        Write [tool.uv].constraint-dependencies in pyproject.toml by merging
+        edx-lint's global constraints with optional repo-specific constraints
+        from [tool.edx_lint].uv_constraints in the same file.
+    """
+    parser = argparse.ArgumentParser(prog="edx_lint write_uv_constraints", add_help=False)
+    parser.add_argument("pyproject", nargs="?", default="pyproject.toml")
+    args, unknown = parser.parse_known_args(argv)
+
+    if unknown:
+        print(f"Unknown arguments: {' '.join(unknown)}")
+        return 1
+
+    global_text = (
+        importlib.resources.files("edx_lint")
+        .joinpath("files/common_constraints.txt")
+        .read_text(encoding="utf-8")
+    )
+    global_constraints, _ = _parse_constraints(global_text)
+
+    try:
+        with open(args.pyproject, encoding="utf-8") as f:
+            data = tomlkit.load(f)
+    except FileNotFoundError:
+        print(f"File not found: {args.pyproject}")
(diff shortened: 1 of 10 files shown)


edx_lint/cmd/write_uv_constraints.py
```diff
@@ -0,0 +1,86 @@
...
+
+import tomlkit
+
+
+def _parse_constraints(text):
+    """Return package-specifier lines from constraints file text, stripping comments."""
+    constraints = []
+    for line in text.splitlines():
+        line = line.split("#")[0].strip()
+        if line and not line.startswith("-"):
```
feanil: We should at least print a warning when we're ignoring lines that start with `-` as those are lines that used to work with the old tooling which will now be ignored.

edx_lint/cmd/write_uv_constraints.py
```diff
@@ -0,0 +1,86 @@
...
+        try:
+            with open(args.local, encoding="utf-8") as f:
+                local_text = f.read()
+            constraints += _parse_constraints(local_text)
+        except FileNotFoundError:
+            print(f"Warning: local constraints file not found: {args.local}")
+
+    # Deduplicate while preserving order (local entries override nothing, but
+    # a package appearing in both files should only appear once).
+    constraints = list(dict.fromkeys(constraints))
```
feanil: Can we just use a set here instead, that seems like it would be clearer.  Also, this doesn't handle the case where the same package is constrained in two different ways right?  We should handle that, if that happens we should prefer the local version of the constraint. The reason is that when testing updates we want to be able to override the global constraints. (We should add a comment to this effect along with the fix.)

feanil:
https://github.com/openedx/openedx-aspects/pull/359 shows what the implementation of this would look like.

edx_lint/cmd/write_uv_constraints.py
```diff
@@ -0,0 +1,113 @@
...
+    )
+    global_constraints, _ = _parse_constraints(global_text)
+
+    local_constraints = []
+    if args.local:
+        try:
+            with open(args.local, encoding="utf-8") as f:
+                local_text = f.read()
+            local_constraints, local_directives = _parse_constraints(local_text)
+            # Warn about pip directives (-c, -r, etc.) that are silently ignored.
```
bmtcril: I do have some concerns about this, since it would be a divergence from what we expect to happen? I think that the overall direction is to move away from these files eventually anyway, right?

bmtcril (approved):
Thanks for taking this on, it was definitely the worst missing piece of the initial uv stab. I just have the one non-blocking concern that we probably can't do anything about.

edx_lint/cmd/write_uv_constraints.py
feanil: Yea, in an ideal world, we could reference global and local constraints directly in the pyproject.toml.  Another alternative that I was considering is instead of adding a file here. We could add a new tool directive section to pyproject.toml for edx_lint so that we could add local constraints there instead and it would combine the local list with the global list.  I opted for the constraints.txt file because it's where people would look to add this but thinking through it again, the tool directive might be a better option, the comment in the pyproject.toml points to the edx-lint section in the same file and because it's not a constraints.txt file, but a new directive, people will be less surprised by it having different capabilities.

What do you think of this alternative approach? Worth pursing? If so I can update the PRs to do that instead before we land it.

bmtcril: Ooh, I think I like that better. If we're biting the bullet on this fairly major update that seems like the more future-proof solution

PR #552, opened 2026-07-07
temp: Constrain social-auth packages for POST issue

timmc-edx:
See https://github.com/openedx/edx-drf-extensions/issues/561

**Merge checklist:**
- [x] If adding new checks, followed how_tos/0001-adding-new-checks.rst
- [x] Changelog record added (if needed)
- [x] Documentation updated (not only docstrings) (if needed)
- [x] Commits are squashed

### DIFF
--- a/edx_lint/files/common_constraints.txt
+++ b/edx_lint/files/common_constraints.txt
@@ -18,3 +18,9 @@ Django<6.0
 # elastic search changelog: https://www.elastic.co/guide/en/enterprise-search/master/release-notes-7.14.0.html
 # See https://github.com/openedx/edx-platform/issues/35126 for more info
 elasticsearch<7.14.0
+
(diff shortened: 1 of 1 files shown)


jansenk (approved):

feanil:
@timmc-edx this change makes sense to mitigate the issue introduced by the upgrade.  But it's not clear to me what the long term solution is to fix this?  I see that in https://github.com/openedx/edx-drf-extensions/issues/561 there is a proposed solution, is anyone actually working on that?

timmc-edx:
No one is currently working on it -- our team is finishing up some edxapp containerization work and will likely not be looking into edx-drf-extensions fixes within a month time frame.

But I also haven't looked into the details enough to assess whether the proposed fixes are what we want, or whether we should instead be pushing back on the social-auth changes, etc.

feanil (approved):

PR #559, opened 2026-07-23
fix: Enhance PII checks to fail on new/existing PII fields in no_pii annotated models

Akanshu-2u:
### Description:
This PR introduces a new pylint checker `(pii-invalid-no-pii-annotation / W7633)` to ensure that Django models explicitly marked as not containing PII remain compliant with [OEP-0030](https://docs.openedx.org/projects/openedx-proposals/en/latest/architectural-decisions/oep-0030-arch-pii-markup-and-auditing.html). It enforces this compliance by failing CI if any model annotated with `.. no_pii:` contains fields that match known PII terms. This ensures that existing incorrect annotations are audited and corrected, and prevents developers from accidentally merging new sensitive fields into these models without proper review and annotation updates.

### Solution:

- **Created a new Pylint plugin:** Added a checker that specifically targets concrete Django models.
- **Annotation validation:** The checker parses model docstrings and preceding comment lines to identify classes explicitly annotated with `.. no_pii:` .
- **Strict field checking:** When a `.. no_pii:` model is found, the checker scans its class attributes and __init__ assignments. If any tentative PII field is detected (matching the configurable pii-terms), it raises a W7633 warning. This forces the developer to either rename the non-sensitive field or update the model's metadata to `.. pii:` .

### PII detection while testing on CI:

- https://github.com/Akanshu-2u/edx-platform/actions/runs/31375567882/job/93413902806

### Inline comment example to bypass linter check (False positives):

```
    # pylint: disable-next=pii-invalid-no-pii-annotation  # field does not store user PII data, safe under OEP-30
    email = models.BooleanField(default=False, null=False, blank=False)
```

### Private JIRA Link:
[BOMS-587](https://2u-internal.atlassian.net/browse/BOMS-587)


edx_lint/pylint/pii_annotation_check.py
```diff
@@ -0,0 +1,288 @@
...
+    def _init_pii_caches(self):
+        """Reset per-module config caches to None."""
+        self._pii_terms_cache = None
+
+    def _ensure_config_cached(self):
+        """Populate pii-terms cache on first call within a module."""
+        if self._pii_terms_cache is not None:
+            return
+        cfg = self.linter.config
+        raw_terms = getattr(cfg, "pii_terms", ["email", "username"])
```
Akanshu-2u: Add None as default and validate if its None then throw an error.

edx_lint/pylint/pii_annotation_check.py
```diff
@@ -0,0 +1,312 @@
...
+    Fires ``pii-invalid-no-pii-annotation`` (W7633) when a concrete Django model
+    is annotated ``.. no_pii:`` but still has fields matching the PII terms list.
+    Abstract and proxy models are skipped, mirroring django_find_annotations scope.
+    """
+
+    name = "pii-annotation-checker"
+
+    # Message definitions
+    msgs = {
+        ("W%d33" % BASE_ID): (
```
ttak-apphelix, line 33: Every annotation-compliance check in annotations_check.py (feature-toggle-needs-doc, toggle-no-name, setting-boolean-default-value, etc.) is registered as an E (error), not W. Since this enforces the same category of thing — OEP-30 compliance intended to fail CI — should this be E7633 for consistency, rather than W7633?

ttak-apphelix, line 33: Can we confirm W7633 doesn't collide with an existing message ID under BASE_ID in another checker (range_check.py, super_check.py, getattr_check.py, etc.)?

edx_lint/pylint/pii_annotation_check.py
```diff
@@ -0,0 +1,312 @@
...
+
+        if not self._is_annotation_eligible_django_model(node):
+            return
+        if not self._class_has_no_pii_annotation(node):
+            return
+
+        pii_fields = self._collect_pii_fields(node)
+        for field_name, field_node in pii_fields:
+            self.add_message(
+                "pii-invalid-no-pii-annotation",
```
ttak-apphelix: Elsewhere in the codebase (TOGGLE_NOT_ANNOTATED_MESSAGE_ID = "feature-toggle-needs-doc"), the message symbol is a class-level constant referenced in both msgs and add_message(). Here "pii-invalid-no-pii-annotation" is inlined as a bare string in both places — minor, but worth aligning with the existing pattern for consistency.

edx_lint/pylint/pii_annotation_check.py
```diff
@@ -0,0 +1,312 @@
...
+_ANNOTATION_LOOKAHEAD = 10
+
+
+def register_checkers(linter):
+    """Register the PII annotation checker."""
+    linter.register_checker(PiiAnnotationChecker(linter))
+
+
+@check_visitors
+class PiiAnnotationChecker(BaseChecker):
```
ttak-apphelix, line 25: annotations_check.py already has AnnotationBaseChecker built on code_annotations.find_static.StaticSearch/AnnotationConfig for parsing .. toggle_name:-style annotations, and .. no_pii:/.. pii: follow the same annotation grammar per OEP-30. Could this checker subclass AnnotationBaseChecker instead of using custom regexes + a 10-line source lookahead? The hand-rolled scan is fragile (e.g. two classes declared close together could bleed a no_pii comment across class boundaries) and duplicates parsing logic this repo already solves correctly elsewhere.

edx_lint/pylint/pii_annotation_check.py
```diff
@@ -0,0 +1,275 @@
...
+                        if self._is_pii_name(target.name):
+                            found.append((target.name, child))
+
+            # Class-level annotated assignment: ``email: str = ""``
+            elif isinstance(child, astroid_nodes.AnnAssign):
+                if isinstance(child.target, astroid_nodes.AssignName):
+                    if self._is_pii_name(child.target.name):
+                        found.append((child.target.name, child))
+
+            # Instance attributes set inside methods: ``self.email = ...``
```
rahulkanneri-2u, line 285: What about `self.email: str = ...`?

edx_lint/pylint/pii_annotation_check.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def _reset_parsed_config(self):
+        """Reset per-module parsed configs to None."""
+        self._parsed_pii_terms = None
+
+    def _parse_and_store_config(self):
+        """Parse pii-terms config on first call within a module."""
+        if self._parsed_pii_terms is not None:
+            return
+        cfg = self.linter.config
```
rahulkanneri-2u: I would recommend a better variable name than `cfg`

edx_lint/pylint/pii_annotation_check.py
```diff
@@ -0,0 +1,275 @@
...
+        self._parse_and_store_config()
+        return self._parsed_pii_terms
+
+    def _is_pii_name(self, name):
+        """
+        Return True if *name* is a PII identifier.
+
+        Substring match of any pii-term inside *name* → PII.
+        """
+        lower = name.lower()
```
rahulkanneri-2u: I would recommend a better variable name than `lower`

edx_lint/pylint/pii_annotation_check.py
```diff
@@ -0,0 +1,275 @@
...
+                            and isinstance(stmt.value, astroid_nodes.Const)
+                            and stmt.value.value is True):
+                        return True
+        return False
+
+    def _class_has_no_pii_annotation(self, node):
+        """
+        Return True if the class docstring carries a ``.. no_pii:`` annotation.
+        """
+        return self._docstring_has_no_pii(node)
```
rahulkanneri-2u: `_class_has_no_pii_annotation` just calls `_docstring_has_no_pii` with no extra logic, So can we just change line number 113 to call `_docstring_has_no_pii` directly and delete this (_class_has_no_pii_annotation) wrapper?

Akanshu-2u: Good suggestion. I inlined the call to `_docstring_has_no_pii(node)` in `visit_classdef` and removed `_class_has_no_pii_annotation` since it had no additional logic.

edx_lint/pylint/pii_annotation_check.py
Akanshu-2u: Updated for readability: renamed `lower` to `normalized_name`.

edx_lint/pylint/pii_annotation_check.py
Akanshu-2u: Renamed `cfg` to `linter_config`.

edx_lint/pylint/pii_annotation_check.py
Akanshu-2u: Good catch. We only handled `self.email = ...` `Assign` in methods; `self.email: str = ...` is `AnnAssign` and was a gap. I’ve added `AnnAssign` handling for instance attrs and a regression test `test_no_pii_annotated_instance_attr_in_method_flagged`.

edx_lint/pylint/pii_annotation_check.py
Akanshu-2u: implemented.

edx_lint/pylint/pii_annotation_check.py
Akanshu-2u: Yes, I confirmed that `W7633` does not collide with any existing message ID under `BASE_ID` across the other checkers (`range_check.py`, `super_check.py`, `getattr_check.py`, etc.).

I’ll keep `W7633` as-is since the `E76xx` series in `annotations_check.py` is reserved for OEP toggle/setting compliance, while `W7630/W7633` belong to the PII checker family.

Akanshu-2u: Answered in this https://github.com/openedx/edx-lint/pull/559#discussion_r3747209887 .

edx_lint/pylint/pii_annotation_check.py
Akanshu-2u: Implemented.

edx_lint/pylint/pii_annotation_check.py
Akanshu-2u: Before the changes, `AnnotationBaseChecker / StaticSearch` was mainly designed for multi-field annotation groups and loads YAML from `code_annotations/contrib/config/`. There is no `pii_annotations.yaml` there.

Since `no_pii` is a standalone annotation rather than an annotation group, `iter_groups()` doesn't cover it. Also, `StaticSearch` only returns the file, line number, and annotation, so we'd still need additional logic to determine which `ClassDef` owns that annotation.

So using it would add complexity without solving the core issue more directly than the current approach.

rahulkanneri-2u (approved):

bmtcril (commented):
Thanks for this, seems reasonable to me after a rebase!

bmtcril (approved):

